PR #1, opened 2015-02-28
String implementation.

mholovaty:
Complete String implementation.


string/string.cc
```diff
@@ -1,12 +1,113 @@
...
+#include <algorithm>
+#include <iterator>
+#include <memory>
+#include <string>
+
+
+const char* EMPTY = "";
+
+
+size_t strlen(const char* s)
```
maciekgajewski: The library provides you with well-tested and super-optimized std::strlen. Do not reinvent the wheel.

string/string.cc
```diff
@@ -1,12 +1,113 @@
...
+	size_t len = 0;
+
+	while (s[len] != '\0')
+		len++;
+
+	return len;
+}
+
+
+String::String() {}
```
maciekgajewski: Or, you can just say:

```
String::String() = default;
```

This seems more explicit to me.

string/string.cc
```diff
@@ -1,12 +1,113 @@
...
+
+String::~String() {}
+
+String::String(const String& o)
+{
+	mSize = o.size();
+
+	mValue = char_ptr(new char[mSize + 1]);
```
maciekgajewski: You can use unique_ptr::reset() here.

string/string.cc
```diff
@@ -1,12 +1,113 @@
...
+
+	std::copy_n(o.begin(), mSize, begin());
+
+	mValue[mSize + 1] = '\0';
```
maciekgajewski, line 29: You may just copy mSize+1 chars. Yoyu are copying from your own String, so you know that it contains the null-terminator.

string/string.cc
```diff
@@ -1,12 +1,113 @@
...
+}
+
+template<typename It> String::String(It first, It last)
 {
-	// TODO your code goes here, if any
+	mSize = std::distance(first, last);
+
+	mValue = char_ptr(new char[mSize + 1]);
+
+	std::copy_n(first, mSize, begin());
```
maciekgajewski: Here you can't assume that the terminating zero is here.
Please note: this constructor, ans the other one share almost all of the code. Maybe you could implement the other ones in terms of this?
(Google: delegating constructor)

string/string.cc
```diff
@@ -1,12 +1,113 @@
...
+}
+
+bool String::operator==(const String& s) const
+{
+	return std::equal(begin(), end(), s.begin());
+}
+
+bool String::operator==(const char* s) const
+{
+	if (empty() and s == '\0') return true;
```
maciekgajewski: Oops! You are comparing _s_ (of type _const char_) to _'\0'_ (of type _char_).

This may work, but this probably not what you wanted to implement. I'm pretty sure your compiler gives a warning here.
You can assume that 's' is not null.

string/string.cc
```diff
@@ -1,12 +1,113 @@
...
+
+	return std::equal(begin(), end(), s);
+}
+
+std::ostream& operator<<(std::ostream& stream, const String& s)
+{
+	for(auto c: s) stream << c;
```
maciekgajewski: Slow, but correct.

string/tests.cc
```diff
@@ -32,68 +32,66 @@ String IntToString(int i)
...
 	std::ostringstream ss;
-	
+
 	ss << s;
-	
+
 	assert(ss.str() == "12345");
 }
 
-// TODO add more tests here. Achieve 100% test coverage
```
maciekgajewski, line 92: TODO: add more tests. Are all methods covered by unit tests?

string/string.cc
maciekgajewski: Moving constructor and assignment operators are missing.

string/string.cc
```diff
@@ -1,12 +1,154 @@
...
+	mSize = std::distance(first, last);
+	mValue.reset(new char[mSize + 1]);
+
+	std::copy_n(first, mSize, begin());
+
+	mValue[mSize + 1] = '\0';
+}
+
+// I don't see any other effective way to concatenate for the given API.
+String::String(const String& a, const String& b)
```
mholovaty, line 20: Is there a smart concatenation without copying each time?
In Java all  '+' are converted to a StringBuffer by compiler. What's the point of '+' if you need to use a buffer anyway?

maciekgajewski, line 20: And string buffer doesn't contain a, uhm, a buffer to which the strings have to be copied?

For 2 strings there is no reasonable optimization. You may want to try to optimize an expression like this:

```
String s = a + b + c + d;
```

by allocating buffer only once. Look for 'expression templates'.

Also, you could optimize operations like this:

```
s += x;
```

by having a pre-allocated buffer in s bigger then needed. This would incur some initial cost, and would allow for avoiding allocation in some cases.

string/string.h
```diff
@@ -117,6 +119,14 @@ class String
 	
 	// Compares with c-string. Assumes s is not null
 	bool operator==(const char* s) const;
+
+protected:
```
maciekgajewski: use 'private'. 'protected' is rarely needed, and even more rarely correct.

string/string.cc
```diff
@@ -1,12 +1,154 @@
...
+String::~String() = default;
+
+String::String(String&& o) = default;
+
+String::String(const String& o): String(o.begin(), o.end()) {}
+
+String::String(const char* s): String(s, s + std::strlen(s)) {}
+
+template String::String(std::string::iterator first, std::string::iterator last);
+
```
maciekgajewski: The definition of function template should be in the header file, so that all clients can 'see' it, and it would be automatically instantiated for all types required.
The the line above, instantiating for std::string::iterator wouldn't be needed.

PR #6, opened 2016-02-10
Complete implementation of linkedlist + listiterator

sytham:


course2/meeting4/MyLinkedList.h
```diff
@@ -0,0 +1,152 @@
...
+template<typename T>
+class ListIterator
+{
+public:
+	//These are necessary in order to use std::copy in line 60
+	typedef T value_type;
+	typedef T* pointer;
+	typedef T& reference;
+	typedef std::ptrdiff_t difference_type;
+	typedef std::forward_iterator_tag iterator_category;
```
maciekgajewski, line 33: The easiest way to get this, is ti inherit std::iterator:

```
template<typename T>
class ListIterator : public std::iterator<std::forward_iterator_tag, T>
{
//...
};
```

course2/meeting4/MyLinkedList.h
```diff
@@ -0,0 +1,152 @@
...
+
+public:
+
+	typedef std::unique_ptr<Node<T>> node_ptr;
+	typedef T value_type;
+
+	MyLinkedList<T>() {}
+
+    MyLinkedList<T>(const MyLinkedList<T>& oldList) {
+    	std::copy(oldList.begin(), oldList.end(), std::front_inserter(*this));
```
maciekgajewski, line 60: Nice, but the resulting list will have reverse order. You should use 'back_inserter', in your list implements push_back.

Usually, container have 'assign' method, that takes two iterators, and populates the content ot the range described by these iterators. If your list had this method, you could call it here.

course2/meeting4/MyLinkedList.h
```diff
@@ -0,0 +1,152 @@
...
+			mFront = mFront->mPrev;
+			mFront->mNext.reset();
+		} else {
+			mBack.reset();
+		}
+		mSize = mSize == 0 ? 0 : mSize-1;
+    }
+
+    const T& front() const {
+    	//I'm gonna stick to std::list API: "Calling front on an empty container is undefined."
```
maciekgajewski, line 89: good choice. Adheres to "don't pay what you are not using" principle

course2/meeting4/MyLinkedList.h
```diff
@@ -0,0 +1,152 @@
...
+	void clear() {
+		mBack.reset();
+		mSize = 0;
+	}
+
+	bool empty() const {
+		return mSize == 0;
+	}
+
+	ListIterator<T> begin() const {
```
maciekgajewski, line 134: This is not entirely correct, as the iterator allows for changing the values of the collection, and the method is 'const'.
Usually containers have iterator types: 'iterator' and 'const iterator'

course2/meeting4/MyLinkedList.h
```diff
@@ -0,0 +1,152 @@
...
+		return ListIterator<T>(mBack.get());
+	}
+
+	ListIterator<T> end() const {
+		return ListIterator<T>(nullptr);
+	}
+
+
+private:
+    //need to use raw pointer for front, otherwise need shared pointer
```
maciekgajewski, line 144: It's fine, no need to explain. This is a valid use of raw pointer.

PR #7, opened 2016-02-21
M10 hw branch

sytham:
Homework after meeting10


course2/meeting10/main.cc
```diff
@@ -0,0 +1,54 @@
...
+#include <functional>
+
+struct RNG
+{
+	std::default_random_engine randEngine;
+	std::uniform_real_distribution<double> realDist{0, 1000};
+	double operator() () { return realDist(randEngine); }
+};
+
+void PrintTime(std::function<void()> func, const char* message)
```
maciekgajewski: std::function is really cool, becuase it erases the type - allows you to squeeze all the different lambdas into one type.
But it also introduces overhead.

In this case, making the function a template would be better:
1. You don't need the type erasure in this case
2. In case of a benchmark, any overhead introduces noise into the result

course2/meeting4/CustomString.h
```diff
@@ -0,0 +1,195 @@
...
+class CustomString
+{
+public:
+	CustomString() {}
+
+	CustomString(const char* str) {
+		copyCharArray(str);
+	}
+
+	int length() { return mChars.size(); }
```
maciekgajewski, line 29: const

course2/meeting4/CustomString.h
```diff
@@ -0,0 +1,195 @@
...
+
+	/* operators on C-style strings */
+	CustomString& operator=(const char* str) {
+		mChars.clear();
+		copyCharArray(str);
+		return *this;
+	}
+
+	bool operator==(const char* str) {
```
maciekgajewski, line 38: const

course2/meeting10/main.cc
```diff
@@ -0,0 +1,54 @@
...
+	PrintTime([&]() {std::accumulate(v.begin(), v.end(), 0);}, "Sum vector: ");
+	PrintTime([&]() {std::accumulate(l.begin(), l.end(), 0);}, "Sum list  : ");
+	PrintTime([&]() {std::accumulate(d.begin(), d.end(), 0);}, "Sum deque : ");
+
+	PrintTime([&]() {std::sort(v.begin(), v.end());}, "Sort vector: ");
+	PrintTime([&]() {l.sort();}, "Sort list  : ");
+	PrintTime([&]() {std::sort(d.begin(), d.end());}, "Sort deque : ");
+
+	PrintTime([&]() {std::nth_element(v.begin(), v.begin() + 100, v.end());}, "Nth element vector: ");
+	PrintTime([&]() {std::nth_element(d.begin(), d.begin() + 100, d.end());}, "Nth element deque : ");
```
maciekgajewski, line 84: I'm curious! What are the results?

course2/meeting10/main.cc
```diff
@@ -0,0 +1,54 @@
...
+	std::cout << message << (end-start).count() << std::endl;
+}
+
+
+int main()
+{
+	int N = 1000000;
+	RNG randomGen;
+
+	std::vector<double> v(N);
```
maciekgajewski: This is bit of a cheating, as you have the containers "pre-grown" before the test. Could either
- start with empty containers, and add elements using push_back
- include the initialization of a container into the benchmark

course2/meeting10/main.cc
```diff
@@ -0,0 +1,54 @@
...
+
+void PrintTime(std::function<void()> func, const char* message)
+{
+	std::chrono::time_point<std::chrono::system_clock> start, end;
+
+	start = std::chrono::system_clock::now();
+	func();
+	end = std::chrono::system_clock::now();
+
+	std::cout << message << (end-start).count() << std::endl;
```
maciekgajewski: I usually display the time like this:

```
using namespace std::literals::chrono_literals;
std::cout << "time: " << (end-start)/1ms << "ms" << std::endl;
```

course2/meeting10/main.cc
```diff
@@ -0,0 +1,54 @@
...
+	std::uniform_real_distribution<double> realDist{0, 1000};
+	double operator() () { return realDist(randEngine); }
+};
+
+void PrintTime(std::function<void()> func, const char* message)
+{
+	std::chrono::time_point<std::chrono::system_clock> start, end;
+
+	start = std::chrono::system_clock::now();
+	func();
```
maciekgajewski: How stable are the result between runs?
You can increase stability by caling func N times in a loop here (N = 10 - 1000), collecting all the times in a vector, and then printing min and median.

course2/meeting10/main.cc
sytham: 1. I'm not sure I understand this. std::function is itself a template class, right? So std::function<void()> means: a function that returns void and takes no arguments, whereas e.g. std::function<int(double,double)> would mean a function that returns int and takes two double arguments? So I don't understand the (practical?) difference with a template.
2. Ok but the benchmark is taken after the overhead is incurred, in this case :-)

sytham: Github erased the template. I meant `std::function<void()>` and `std::function<int(double,double)>`

course2/meeting10/main.cc
sytham: Yeah you're right :-) There currently isn't much difference between the three for populating hte containers -- that'll probably be quite different if init is included in the benchmark

course2/meeting4/CustomString.h
sytham, line 29: Whoops didn't mean to include this (and the files below) in the pull request. It's only about main.cc

course2/meeting10/main.cc
sytham, line 84: Populate vector: 28071929
Populate list  : 62876913
Populate deque : 22256856
Sum vector: 124
Sum list  : 4063306
Sum deque : 778919
Sort vector: 99146610
Sort list  : 358063715
Sort deque : 123160067
Nth element vector: 2786591
Nth element deque : 3312334

This is with container init as part of the populate container benchmark. These numbers are quite different from the ones on my home computer though (I compiled and ran this on www.cpp.sh); especially vector sum seems unbelievably small

course2/meeting10/main.cc
maciekgajewski: std::function is a class that _wraps_ any type of functor. Creating, copying and calling it is slower than calling an unwrapped functor.
So your benchmark also measures the call overhead

course2/meeting10/main.cc
maciekgajewski, line 84: Time-based benchmarks should be run on a bare-metal machine, otherwise the results are very noisy.
If you happen to run it on physical machine, please, send me the results, I'll share them with the rest of the group.

PR #8, opened 2016-04-09
Maurice-homework

zerokill:
Added a print statement for the polar coordinates. One question about the stream overloading operator: I tried to make this a member class but the compiler (clang) kept complaining about

`error: invalid operands to binary expression ('ostream' (aka 'basic_ostream<char>') and 'Vector')`

This is the member method:

`std::ostream& operator<<(std::ostream& out) {`
`return out << "{x= " << this->GetX() << " , y= " << this->GetY() << "}" << std::end;`
`}`

Google says something that I should use a friend attribute but its not completely clear why I can't overload it in the same way as the other operators.

See you next week!

### DIFF
--- a/course3/homework-vector/vector.cc
+++ b/course3/homework-vector/vector.cc
@@ -1,10 +1,44 @@
 #include <iostream>
 #include <cassert>
+#include <cmath>
 
 class Vector
 {
-    // TODO
+    int mX, mY;
+
+public:
+    Vector() {}
+    Vector(int x, int y) : mX(x), mY(y) {}
+
+    int GetX() const { return mX; }
+    int GetY() const { return mY; }
+
+    double GetPolarLength() const { return sqrt((mX * mX) + (mY * mY)); }
+    double GetPolarAngle() const { return atan2(mX, mY); }
+
+    void SetX(int x) { mX = x; }
+    void SetY(int y) { mY = y; }
+
+    bool operator==(const Vector& rhs) {
+        return ((rhs.GetX() == this->GetX()) && (rhs.GetY() == this->GetY()));
+    }
+
+    bool operator!=(const Vector& rhs) {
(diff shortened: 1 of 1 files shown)


maciekgajewski:
The method form of operator overloading would expect "this" (in our case, the Vector) to be the left-hand side operand. Wee need vector to be the rhs of the expression, hence the free-standing form.

See more here: http://en.cppreference.com/w/cpp/language/operators

course3/homework-vector/vector.cc
```diff
@@ -1,11 +1,45 @@
...
+
+    bool operator==(const Vector& rhs) {
+        return ((rhs.GetX() == this->GetX()) && (rhs.GetY() == this->GetY()));
+    }
+
+    bool operator!=(const Vector& rhs) {
+        return !((rhs.GetX() == this->GetX()) && (rhs.GetY() == this->GetY()));
+    }
+
+    Vector operator+(const Vector& rhs) {
```
maciekgajewski, line 30: Q: can this operators be 'const' methods? Should they be?

course3/homework-vector/vector.cc
```diff
@@ -1,11 +1,45 @@
...
 #include <cassert>
+#include <cmath>
 
 class Vector
 {
-    // TODO
+    int mX, mY;
+
+public:
+    Vector() {}
```
maciekgajewski, line 10: What are the  values of mX and mY after default initializtion?

zerokill, line 10: They are uninitialized and because of that undefined

course3/homework-vector/vector.cc
zerokill, line 30: This method can be const because the plus operator is not changing anything in the arguments of in the class. It's creating a new class with the summed X,Y

PR #9, opened 2016-04-11
Homework exercise one

dgkimpton:
Homework exercise one

### DIFF
--- a/course3/meeting8/meeting8.cc
+++ b/course3/meeting8/meeting8.cc
@@ -21,49 +21,21 @@ void print(const Range& range)
 }
 
 
+struct S
+{
+	int f;
+	S(int x) : f(x) {}
+	S() : f(7) {}
+};
+
 int main(int argc, char** argv)
 {
 	std::cout << "Hello" << std::endl;
-	auto x = argc + argv;
-	print_range(argv, x);
+	std::array<S, 5> a;
 	
-	std::vector<std::string> strings;
-	strings.push_back("One");
-	strings.push_back("Two");
-	strings.push_back("Three");
-	strings.push_back("Four");
-	
-	print_range(strings.begin() + 2, strings.begin()+4);
-
-	std::list<int> ints;
-	ints.push_back(4);
-	ints.push_back(7);
-	
-	print_range(ints.begin(), ints.end());
-	
-	std::deque<double> dbls;
-	dbls.push_back(3.14);
-	dbls.push_back(77.77);
-
-	print(dbls);
-	
-	std::array<bool, 3> bools;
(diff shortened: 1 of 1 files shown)


course3/meeting2/hello.cc
```diff
@@ -1,13 +1,56 @@
...
+}
 
 int main(int argc, char** argv)
 {
-    std::string name;
+//  std::cout << "Hi There" << std::endl;
+//  SimpleString s("Hi Back");
+//  std::cout << s << std::endl;
+
+    SimpleString name;
```
maciekgajewski: Have you tried printing default-constructed string?

course3/meeting2/hello.cc
```diff
@@ -1,13 +1,56 @@
...
+    SimpleString()
+    {
+    }
+
+    SimpleString(const char* ascii)
+    {
+      int lastCharacterIndex = BUFFER_SIZE - 1;
+
+      int i = 0;
+      while(i < lastCharacterIndex
```
maciekgajewski: The would be sightly better as a "for" loop, and much better as a call to std::strcpy

maciekgajewski:
And what about the Vector exercise?

course3/meeting2/hello.cc
```diff
@@ -1,13 +1,56 @@
...
+            && ascii[i] != 0)
+      {
+        mBuffer[i] = ascii[i];
+	++i;
+      }
+      mBuffer[i] = 0;
+    }
```
maciekgajewski: Is the type copyable? Is it assignable? Have you tried compiling the following code:

```
SimpleString a("hello");
SimpleString b = a;
```

or

```
SimpleString a("hello");
SimpleString b;
b = a;
```

course3/meeting2/hello.cc
dgkimpton: doh, it actually worked on my system by random chance, but indeed, will fix.

course3/meeting2/hello.cc
dgkimpton: std::strcpy I can see, but why is for better than while in this case?

int i;
      for (i=0; i < lastCharacterIndex && ascii[i] != 0; ++i)
      {
        mBuffer[i] = ascii[i];
      }
      mBuffer[i] = 0;

maciekgajewski: Because 'for' is canonical for of this kind of loop. More obvious to the reader;. Still, std::strcpy is _the_ cleanest solution.

dgkimpton:
I hadn't spotted the Vector exercise initially, I think it is completed now.

PR #10, opened 2016-04-11
Added Vector class definition.

wiggity:
It compiles therefore it works :)

### DIFF
--- a/course3/homework-string/my_string.h
+++ b/course3/homework-string/my_string.h
@@ -1,19 +1,52 @@
 #pragma once
 #include <iostream>
 #include <memory>
+#include <cstring>
 
 class MyString
 {
-    private:
-        //unique_ptr<char*> buf;
+	private:
+		std::unique_ptr<char[]> buf;
 
-    public:
-        MyString(const char* my_string) {
-            std::cout << "Constructor" << std::endl;
-            //buf = new char[strlen(my_string)];;
-        }
+	public:
+		MyString(const char* my_string) {
+			std::cout << "String literal constructor" << std::endl;
+			buf = std::make_unique<char[]>(std::strlen(my_string));
+			strcpy(buf.get(), my_string);
+		}
 
-        ~MyString() {
-            std::cout << "Destructor" << std::endl;
-        }
+		MyString() {
+			std::cout << "Default constructor" << std::endl;
+			buf = std::make_unique<char[]>(0);
+		}
+
+		MyString(const MyString& obj){
+			std::cout << "Copy constructor" << std::endl;
+			buf = std::make_unique<char[]>(obj.size());
+			strcpy(buf.get(), obj.buf.get());
+		}
+
+		~MyString() {
+			std::cout << "Destructor" << std::endl;
+		}
+
+		bool operator== (const MyString& rhs) const {
+			return std::strcmp(buf.get(), rhs.buf.get()) == 0;
+		}
+
+		const char& operator[] (int i) const {
+			return buf.get()[i];
+		}
+
+		char& operator[] (int i) {
+			return buf.get()[i];
+		}
+
+		bool empty() const {
+			return std::strlen(buf.get()) == 0;
+		}
+
+		int size() const {
+			return std::strlen(buf.get());
+		}
 };
--- a/course3/homework-string/my_string.h
+++ b/course3/homework-string/my_string.h
@@ -9,14 +9,14 @@ class MyString
 		std::unique_ptr<char[]> buf;
 
 	public:
-		MyString() : buf(std::make_unique<char[]>(0)) {}
+		MyString() = default;
 
-		MyString(const char* str) : buf(std::make_unique<char[]>(std::strlen(str))) {
-			strcpy(buf.get(), str);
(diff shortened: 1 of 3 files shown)


maciekgajewski:
Good, modulo const-correctness. I know we haven't discussed that one yet :)

course3/homework-vector/vector.cc
```diff
@@ -1,9 +1,55 @@
...
-    // TODO
+    private:
+        int private_x, private_y;
+
+    public:
+        // required constructors
+        Vector(int x, int y): private_x(x), private_y(y) {}
+        Vector() {
+            private_x = 0;
+            private_y = 0;
```
maciekgajewski: Look what's happening here: the member variables are first _default initialized_, and then you _assign_ to them.
I case of integer the first one is a no-op, but just for the sake of excercise: can you do it w/o assignment?

Also: can you make the object default-constructible with known default values of the fields without implementing the default constructor at all?

course3/homework-vector/vector.cc
```diff
@@ -1,9 +1,55 @@
...
+            if (private_x == rhs.GetX() && private_y == rhs.GetY())
+                return true;
+            return false;
+        }
+
+        bool operator!= (Vector rhs) {
+            return !operator==(rhs);
+        }
+
+        Vector operator+ (Vector rhs) {
```
maciekgajewski: I all the functions here, you take the parameter by value, which means deep-copying. In this case it's just 2 its just 2 ints, so no biggie, but for the sake of exercise - can you change the signatures so that the values are passed by reference instead of copy?

wiggity:
Passing args by reference instead of by value now.

course3/homework-vector/vector.cc
```diff
@@ -1,9 +1,55 @@
...
+
+        float GetTheta() {
+            return std::atan2(private_y, private_x) * 180 / M_PI;
+        }
+
+        float GetR() {
+            return std::pow(std::pow(private_x, 2) + std::pow(private_y, 2), 0.5);
+        }
+
+        friend std::ostream& operator<< (std::ostream& stream, Vector& v) {
```
maciekgajewski: Now, try compiling this code:

```
std::cout << "The sum is: " << (v1+v2) << std::endl;
```

wiggity:
Using the compiler's generated default constructor instead of my own.

course3/homework-vector/vector.cc
```diff
@@ -1,9 +1,53 @@
...
 {
-    // TODO
+    private:
+        int private_x;
+        int private_y;
+
+    public:
+        // required constructors
+        Vector(int x, int y): private_x(x), private_y(y) {}
+        Vector() = default;
```
maciekgajewski, line 14: How will this initialize the members? what the values of private_x and private_y will be in default-constructed object?

wiggity, line 14: The value of a default constructed int is 0.

wiggity, line 14: The members will get initialized as pat of the default constructor.  An integer will be default initialized to 0.

maciekgajewski, line 14: The value of default-initialized int is unspecified. There is something called "zero initialization", but this is not what's happening here.
http://en.cppreference.com/w/cpp/language/default_initialization

wiggity, line 14: Fixed.

course3/homework-vector/vector.cc
wiggity: Fixed by passing the arg by value rather than by reference.

maciekgajewski: Is it really fixed? What if the object would be expensive to copy?

wiggity:
Alright, I've given the string class my best shot.

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,45 @@
...
+#include <cstring>
 
 class MyString
 {
-	// TODO
+	private:
+		std::unique_ptr<char[]> buf;
+
+	public:
+		MyString() : buf(std::make_unique<char[]>(0)) {}
```
maciekgajewski: Do not allocate 0 bytes. This may be an undefined behaviour.

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,45 @@
...
+
+		MyString(const char* str) : buf(std::make_unique<char[]>(std::strlen(str))) {
```
maciekgajewski: Q: what value does `std::strlen("abc")` returns? 
Q: how many bytes do you need to store `"abc"`?

maciekgajewski:
Please, run the string test trough valgrind, it should show some warnings

course3/homework-string/my_string.h
maciekgajewski: So this is how you use `make_unique` with arrays! I learned something.

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,45 @@
...
+			strcpy(buf.get(), str);
```
maciekgajewski: Style: if you use `std::strlen`, use `std::strcpy`

wiggity:
Updated the code after reviewing your comments.

PR #12, opened 2016-04-20
Vector class implementation

igorustych:
Sorry for the late submission, I just returned from a holiday so this is literally the first evening I'm at home.

### DIFF
--- a/course3/homework-vector/vector.cc
+++ b/course3/homework-vector/vector.cc
@@ -1,33 +1,89 @@
 #include <iostream>
 #include <cassert>
+#include <iostream>
+using namespace std;
 
 class Vector
 {
-    // TODO
+	public:
+
+		Vector()
+			: _x{}, _y{}
+		{}
+
+		Vector(int initial_x, int initial_y)
+			: _x{initial_x}, _y{initial_y}
+		{}
+
+		int GetX() const
+		{
+			return this->_x;
+		}
+
+		void SetX(int new_x)
+		{
+			this->_x = new_x;
+		}
+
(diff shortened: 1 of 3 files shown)


course3/homework-vector/vector.cc
```diff
@@ -1,33 +1,89 @@
...
+#include <iostream>
+using namespace std;
 
 class Vector
 {
-    // TODO
+	public:
+
+		Vector()
+			: _x{}, _y{}
```
maciekgajewski, line 11: what will the values of x and y be?

course3/homework-vector/vector.cc
```diff
@@ -1,33 +1,89 @@
...
+		{
+			return this->_y;
+		}
+
+		void SetY(int new_y)
+		{
+			this->_y = new_y;
+		}
+
+		inline bool operator==(const Vector& other)
```
maciekgajewski, line 38: 'inline' is implied for methods defined in class body, so it's redundant here.

course3/homework-vector/vector.cc
```diff
@@ -1,33 +1,89 @@
...
+			return !(*this == other);
+		}
+
+		inline Vector operator+(const Vector& other)
+		{
+			Vector vector(this->GetX() + other.GetX(), this->GetY() + other.GetY());
+			return vector;
+		}
+
+		friend ostream& operator<<(ostream& os, const Vector& v)
```
maciekgajewski, line 54: No need to make it friend, if you are not accessing any private fields

PR #14, opened 2016-05-08
Homework string mau

zerokill:
Done with homework. I am not quite sure about the amount of c code in there. eg determining the length of the string using pointers.

Or was I supposed to use c++ arrays instead of c arrays?

Regards,

Maurice

### DIFF
--- a/course3/homework-string/my_string.h
+++ b/course3/homework-string/my_string.h
@@ -16,15 +16,17 @@ public:
     MyString(const char *data) {
         length = std::strlen(data);
 
-        buff = std::make_unique<char []>(length);        
+        buff = std::make_unique<char []>(length + 1);        
 
         std::strcpy(buff.get(), data);
     }
 
     MyString(const MyString& rhs) {
         length = rhs.size();
-        buff = std::make_unique<char []>(length);        
-        std::strcpy(buff.get(), rhs.buff.get());
+        if (rhs.buff.get()) {
+            buff = std::make_unique<char []>(length + 1);        
+            std::strcpy(buff.get(), rhs.buff.get());
+        }
     }
 
     bool empty() const {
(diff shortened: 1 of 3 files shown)


course3/homework-string/my_string.h
```diff
@@ -2,5 +2,73 @@
...
+
+    bool empty() {
+        return length == 0 ? true : false;
+    }
+
+    int size() const {
+        return length;
+    }
+
+    int size() {
```
maciekgajewski: what's the use of non-const size()?

course3/homework-string/my_string.h
```diff
@@ -2,5 +2,73 @@
...
+
+public:
+    MyString() {
+        length = 0;
+    }
+
+    MyString(const char *data) {
+        length = strlen(data);
+
+        buff = new char[length];
```
maciekgajewski: Where do you free the memory?

maciekgajewski:
There is way too much C indeed. If you re-read my tips, you'll notice that I recommend the use of std::strlen and std::strcmp functions to measure and compare the string.
Also, you allocate memory, but you never delete it. Run your program trough valgrind, and you'll see!

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,53 @@
...
+    int size() const {
+        return length;
+    }
+
+    bool operator!=(const MyString& rhs) {
+        return !operator==(rhs);
+    }
+
+    bool operator==(const MyString& rhs) {
+        return std::strcmp(buff.get(), rhs.buff.get()) == 0;
```
zerokill: Question about this: How am i able to access a private of rhs? Does the compiler recognise that this function is called in the class method of this private?

PR #16, opened 2016-05-10
string and vector homeworks

karunrao97:
Some questions on const:
1. The code passes all tests whether or not I use const in function arguments. For example, both "char& operator[](const int idx)" and "char& operator[](int idx)" work fine, as do "MyString(const MyString& str)" and "MyString(MyString& str)". Any reason to prefer one over the other in either case?
2. Similar to above, the code passes all tests whether or not I use const in return types. For example, both "const bool empty() const" and "bool empty() const" work fine, as do "const char& operator[](const int idx) const" and "char& operator[](const int idx) const". Again, any reason to prefer one over the other in either case?


maciekgajewski:
'const' has little meaning with simple types (int, bool). It is more important in case of pointers and references, because then it means that the pointed-to objects can't be modified, and only const methods of the object can be called.

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,64 @@
...
+#include <memory>
+
 class MyString
 {
-	// TODO
+private:
+	std::unique_ptr<char[]> mBuffer;
+
+public:
+	MyString(): mBuffer(new char[1]) {}
```
maciekgajewski: 1. you should explicitly init the last (and first here) character to 0, otherwise it is left uninitialized.
2. nice container should not do any allocation in default constructor. Allocation may be expensive, so why do it to store nothing?

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,64 @@
...
+
+	MyString(const char* str): mBuffer(new char[std::strlen(str)])
```
maciekgajewski: Question: what would `std::strlen("abc")` return? How many bytes do you need to store `"abc"` ?

PR #17, opened 2016-05-10
please review my code maciek

maro243:


course3/homework-string/my_string.h
```diff
@@ -1,6 +1,63 @@
...
+	char& operator[](const int); // indexing operator
+	char& operator[](const int) const; // const indexing operator
+};
+
+MyString::MyString():MyString(""){};
+
+MyString::MyString(const char* inchar){
+	int n = std::strlen(inchar);
+	data = std::unique_ptr<char[]>(new char[n]);
+ 	std::strcpy(data.get(), inchar);
```
maciekgajewski: You forgot about the zero-terminator in the string. `"abc"` requires 4-character buffer: a, b, c, \0, but `std::strlen("abc")` will return 3. If you decide to store null-terminated string in your buffer, you need to remember to add one character to store the terminator.

`stdcpy` copies the string and null-terminator. Because your buffer is too small, it writes over the memory you don't own. It should show-up in valgrind.

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,63 @@
...
+	bool empty() const;
+	int size() const;
+
+	friend bool operator== (const MyString &s1, const MyString &s2); // equality operator is friended
+	friend bool operator!=  (const MyString &s1, const MyString &s2);
+	char& operator[](const int); // indexing operator
+	char& operator[](const int) const; // const indexing operator
+};
+
+MyString::MyString():MyString(""){};
```
maciekgajewski: This may be correct, but you perform unnecessary allocation only store empty string.

course3/homework-string/my_string.h
maciekgajewski: Because you fail to add 1 to the buffer size, interesting things are happening with empty strings - you allocate 0 bytes.

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,63 @@
...
+}
+
+char& MyString::operator[](const int i){
+	return data.get()[i];
+}
+
+char& MyString::operator[](const int i) const{
+	char a = data.get()[i];
+	char &ref_a = a;
+	return ref_a;
```
maciekgajewski: whoa! You are returning a pointer (reference) to local variable. The variable is no longer valid when the function returns. Any access to the reference is a violation, and should be reported by valgrind.

I know why you did it; it has to do with const-correctness. Please, compare the signature of your operator[] with the ones of `std::string`

maciekgajewski:
Please, run the program trough valgrind, or another similar tool. There is few invalid operations there.

maro243:
Hi Maciek, I've implemented most of the methods without any memory leaks.
Sorry for being a bit behind with homework but I will get through everything at my own pace, I am enjoying the course!

course3/homework-string/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -std=c++11")
```
maciekgajewski: We use C++14

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,113 @@
+#include "my_string.h"
+# include <iostream>
+
+MyString::MyString(){
+//	std::cout << "default constructor" << std::endl;
```
maciekgajewski: Remove the comment-out code once you don't use it anymore

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,113 @@
...
+	std::strcpy(data.get(), that.data.get());
+}
+
+
+bool MyString::empty() const{
+	return size() == 0;
+};
+
+int MyString::size() const{
+	if (data == 0)
```
maciekgajewski: Please, don't compare pointer to int literal, this makes code confusing ("is 'data' an integer?). Either compare it to `nullptr`, or just take advantage of implicit boolean conversion.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,113 @@
...
+		return 0;
+	else
+		return std::strlen(data.get());
+};
+
+bool operator==(const MyString &s1, const MyString &s2){
+//	std::cout << "comparison operator " << s1.data.get()<< "==" << s2.data.get() << "?" << std::endl;
+	return std::strcmp(s1.data.get(), s2.data.get()) == 0;
```
maciekgajewski: What if data is null on any of the strings?

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,113 @@
...
+char& MyString::operator[](const int i){
+	return data.get()[i];
+}
+
+char MyString::operator[](const int i) const{
+	return data.get()[i];
+}
+
+const char* MyString::c_str() const{
+	return data.get();
```
maciekgajewski: What if data is null?

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,113 @@
...
+MyString& MyString::operator+=(const MyString& s2){
+	MyString new_str= *this + s2;
+	data = std::move(new_str.data);
+	return *this;
+}
+
+bool MyString::operator<(const MyString& s2){
+	if (this->empty()and s2.empty()){return false;}
+	if (this->empty()and !s2.empty()){return true;}
+	if (!this->empty()and s2.empty()){return false;}
```
maciekgajewski: If your `c_str()` worked correctly you could reduce it to single strcmp call

course3/homework-string/my_string.cc
maro243: what should the desired behaviour be, return std::nullptr or const char\* "" ?

course3/homework-string/CMakeLists.txt
maro243: yeah I'm using gcc 4.8.4 (comes with ubuntu 14), I only discovered last night that the right flag for c++14 is -std=c++1y

course3/homework-uniqueptr/test.cc
```diff
@@ -0,0 +1,20 @@
...
+	std::strcpy(a.get(),"abcd");
+	std::cout << "value of a=" << a.get() <<std::endl;	
+}
+
+int main(){
+	std::cout<<"test std::unique_ptr"<<std::endl;
+	test<std::unique_ptr<char[]>>();
+	std::cout<<std::endl;
+	std::cout<<"test MyUniquePtr"<<std::endl;
+	test<MyUniquePtr<char[]>>();
```
maro243, line 19: Hi Maciek, looking for some advice

at compile time I'm getting the following which works for std::unique_ptr
test.cc:9:23: error: no matching function for call to ‘MyUniquePtr<char []>::MyUniquePtr(char_)
The type I instantiate MyUniquePtr with is char [],  the call to (new char[5]) returns type char \* so I see why this is happening, but I can't think of any solution, I would hope that the compiler could kind of ignore the difference between char_ and char[]. My suspicion is maybe there is no neat solution for arrays and I would need to add specialisation for that case, I think I remember you saying something about that.

On another note, I was trying to structure my test function to just take the class template MyUniquePtr, and delved into the concept of class template templates. Couldn't really get it working so I just decided to pass the instantiated UniquePtr<char[]> to the test function template instead. Would you be able to off any advice on how to set up tests for template classes?

course3/homework-string/my_string.cc
maciekgajewski: Follow the std::strings's behaviour:
http://en.cppreference.com/w/cpp/string/basic_string/c_str

course3/homework-uniqueptr/test.cc
maciekgajewski, line 19: 1. Don't count on compiler ignoring the difference between `char` and `char[]`. They are _always_ different types, and, sadly, in different contexts the [] means different thing.
   std::unique_ptr provides _specialization_ for [] types, which stores arrays instead of single objects. If you don't know how to create such a specialization, please, look for inspiration in the code of std::unique_ptr.

Please note that unique_ptr<cha> and unique_ptr<char[]> are completely different types, with different interface, that do different things.
1. Template templates are nightmare fuel. The solution you used is a valid one.

PR #18, opened 2016-05-11
Igor's string class implementation

igorustych:
Igor's string homework.

### DIFF
--- a/course3/homework-string/my_string.h
+++ b/course3/homework-string/my_string.h
@@ -1,6 +1,60 @@
 #pragma once
+#include <vector>
+#include <iostream>
+#include <cstring>
 
 class MyString
 {
-	// TODO
+public:
+	MyString()	
+	{ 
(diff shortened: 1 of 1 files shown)


course3/homework-string/my_string.h
```diff
@@ -1,6 +1,60 @@
...
+		return innerStorage[index];
+	}
+	
+	char operator[](int index) const
+	{
+		return innerStorage[index];
+	}
+
+private:
+	std::vector<char> innerStorage;
```
maciekgajewski, line 58: This will, of course, work.
The goal of this exercise was to play with memory management. If you re-read my email, I specifically asked for explicit memory management, and to use unique_ptr<char[]>. Otherwise the task is too easy and you'll learn nothing.

maciekgajewski:
Cheating! The goal was to use unique_ptr

PR #19, opened 2016-05-16
Willis MyString

wiggity:
Closed my old pull request which combined the first string assignment along with the older vector assignment.  Made this branch from your last commit and cherry picked my existing string code into this branch. Added assignment and concatenation operators. Added cmake support.


course3/homework-string/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_FLAGS "-g -Wall")
```
maciekgajewski: Don't set '-g' in your CMake.
To choose whether you want to build in debug/release or any the of the available modes, use -DCMAKE_BUILD_TYPE=XXX when running cmake for the first time.
This way you can have multiple build directories with different configuration.

You only need to set the var on the first run, it will be cached for subsequent runs

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,63 @@
...
+}
+
+MyString::MyString(const MyString& obj) : buf(std::make_unique<char[]>(obj.size()+1)) {
+	std::strcpy(buf.get(), obj.buf.get());
+}
+
+bool MyString::operator==(const MyString& rhs) const {
+	if (!rhs.buf xor !buf)
+		return false;
+	if (!rhs.buf and !buf)
```
maciekgajewski: I strongly advise against using the text form of logical operators.
Not all compilers support them (_cough_VS_cough_), and they are bitwise operatrors, not boolean operators (`&` vs `&&`), so they can produce surprising results at times.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,63 @@
...
+	return std::strcmp(buf.get(), rhs.buf.get()) == 0;
+}
+
+bool MyString::operator!=(const MyString& rhs) const {
+	return !operator==(rhs);
+}
+
+const char& MyString::operator[](int i) const {
+	if (!buf)
+		throw std::range_error("Trying to index unallocated buffer.");
```
maciekgajewski, line 33: This isn't entirely valid.
Either you decide that your `[]` does boundary checks, and then you check both the existence of buffer, and the bounds of the index (which would be easier if index is if unsigned type), or you don't do any checks at all.

Please note: the standard containers (`std::string`, `std::vector`) don't perform any checks on `operator[]`, but offer alternative function, `at(size_t)` that does the checks.

Remember the "don't pay for what your not using" principle!

course3/homework-string/CMakeLists.txt
wiggity: Fixed.

course3/homework-string/my_string.cc
wiggity: Thanks for that. Fixed.

course3/homework-string/my_string.cc
wiggity, line 33: Fixed.

PR #20, opened 2016-05-16
Homework string

neuraltoys:
I have added an mSize member to keep the size
Then I have also added the indexing operator and the not equal comparison operator
now all the tests are fine

I'm not sure if I was supposed to have two different indexing operators, one read-only and another one with write access, but I guess we can discuss this in the next meeting

Alex


neuraltoys:
pull request to wrong branch

PR #21, opened 2016-05-16
String Homework Alex

neuraltoys:
I have added an mSize member to keep the size
Then I have also added the indexing operator and the not equal comparison operator
now all the tests are fine

I'm not sure if I was supposed to have two different indexing operators, one read-only and another one with write access, but I guess we can discuss this in the next meeting

Alex

### DIFF
--- a/course3/homework-string/my_string.cc
+++ b/course3/homework-string/my_string.cc
@@ -1,46 +1,58 @@
+#include "stdafx.h"
 #include "my_string.h"
 
 #include <cstring>
 
-namespace Maciek{
-	
-String::String(const char *inchar)
-    :mBuffer(new char[std::strlen(inchar) + 1])
-{
-	std::strcpy(mBuffer.get(), inchar);
-}
+namespace Maciek {
 
-String::String(const String& str)
-{
-	if (!str.empty())
+	String::String(): mBuffer(nullptr), mSize(0) { }
+
+	String::String(const char *inchar)
+		:mSize(std::strlen(inchar)), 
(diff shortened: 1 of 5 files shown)


course3/homework-string/my_string.cc
```diff
@@ -1,46 +1,58 @@
...
-{
-	std::strcpy(mBuffer.get(), inchar);
-}
-
-String::String(const String& str)
-{
-	if (!str.empty())
+namespace Maciek {
+
+	String::String(): mBuffer(nullptr), mSize(0) { }
```
maciekgajewski, line 8: no need to initialize mBuffer. It will be set to null in its default constructor.

course3/homework-string/my_string.cc
```diff
@@ -1,46 +1,58 @@
...
+	bool String::operator!=(const String& other)
+	{
+		return !(*this == other);
+	}
+
+	char& String::operator[](const int i) const
+	{
+		if (i > mSize)
+		{
+			throw new std::exception();
```
maciekgajewski, line 53: This is not how you throw exceptions!
Or: this is not how you correctly throw exceptions. Besides, I don't think it would compile on standard-compliant compilers. I'm guessing you're using Visual Studio?

course3/homework-string/test.cc
```diff
@@ -1,3 +1,4 @@
+#include "stdafx.h"
```
maciekgajewski, line 1: I wouldn't put this into the code, this is some VS stuff. I would try to make it a standard C++

PR #22, opened 2016-05-17
MyString-1.0.1

dancollier:


course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,64 @@
+#include <iostream>
+#include <cstring>
+#include "my_string.h"
+
+MyString::MyString(){
+	buffer = std::unique_ptr<char[]>(new char[1]);
+	buffer = '\0';
```
maciekgajewski, line 7: Here yu are setting the _pointer_ to null, dallocating the 1-char buffer allocated just above.
If you'd compile with warning enabled, compiler should warn you here.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,64 @@
...
+	int i;
+	for (i = 0; i < std::strlen(theString.buffer.get()); i++){
+		buffer[i] = theString[i];
+	}
+	buffer[i] = '\0';
+}
+   
+bool MyString::empty() const {
+	if (buffer){
+		return buffer[0] == '\0' ? true : false ;
```
maciekgajewski, line 30: Why the `?:`. Why not just `return buffer[0] == '\0'` ?

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,64 @@
...
+size_t MyString::size() const {
+	if (buffer){
+		return std::strlen(buffer.get());
+	}
+	else return 0;
+}	
+
+const bool MyString::operator == (const MyString& theString){
+	if (buffer && theString.buffer){
+		return std::strcmp(buffer.get(), theString.buffer.get()) == 0 ? true : false ;
```
maciekgajewski, line 44: again - unneccesarty `?:`

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,64 @@
...
+	}
+	else if ( buffer == nullptr && theString.buffer == nullptr ){
+		return true;
+	}
+	else return false;
+}
+
+const bool MyString::operator != (const MyString& theString){
+	if ( buffer && theString.buffer){
```
maciekgajewski, line 53: wouldn't it be easier to just call the operator== ans return inverted value?

course3/homework-string/my_string.cc
dancollier, line 7: I am trying to assign the null-termination character to the default MyString character.

I did this because if I didn't, valgrind complained about the default constructor and the size() method: Conditional jump or move depends on uninitialised value(s)

dancollier, line 7: I also tried:
buffer = std::unique_ptr<char[]>(new char('\0'));

But valgrind complained about: Mismatched free() / delete / delete [] 
and even creating a destructor didn't make that warning go away.

As written, the only warning in my compiler is regarding for loops:
  my_string.cc:14:16: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]   for (i = 0; i < std::strlen(theString); i++){

maciekgajewski, line 7: this assign 0 to the _pointer itself_, essentially making it a null pointer:

```
buffer = '\0';
```

What you wanted to do, is to assign 0 to _the first character of the buffer pointed by the pointer_:

```
buffer[0] = '\0';
```

maciekgajewski, line 7: `buffer = std::unique_ptr(new char('\0'));`
will fail, because there are two types of allocation: single object allocation (new/delete), and array allocation (new[N]/delete[]).
This form of unique pointer you use here (`std::unique_ptr<char[]>`) works with array allocation, while `new char` allocates single  char, so it's a single object allocation.

PR #23, opened 2016-05-17
Implemented c_str(), operator=, operator+, operator+=

karunrao97:


maciekgajewski:
Correct!

course3/homework-string/my_string.cc
```diff
@@ -53,3 +88,18 @@ char& MyString::operator[](int idx)
...
+	MyString retStr;
+	retStr.mBuffer.reset(new char[size() + str.size() + 1]);
+	std::strcpy(retStr.mBuffer.get(), mBuffer.get());
+	std::strcpy(retStr.mBuffer.get() + size(), str.mBuffer.get());
+	return retStr;
+}
+
+MyString& MyString::operator+=(const MyString& str)
+{
+	*this = *this + str;
```
karunrao97, line 91: Are we doing unnecessary memory allocation here? It looks like " *this + str " creates a temporary object (let's call it tmpStr) and allocates a new char array in the process, then " *this = tmpStr " copies tmpStr into *this using the copy assignment operator, which allocates yet another array, even though we could theoretically just take over the array from tmpStr since tmpStr is no longer required.

I tried the following, which passes all test, but looks much uglier (also not sure if it's safe):
std::unique_ptr<char[]> tmpBuffer = (*this + str).mBuffer;
mBuffer = std::move(tmpBuffer);
return *this;

Can we do better? The following also passes tests, but I suspect it's also unnecessarily allocating a new char array twice:
*this = std::move(*this + str);
return this;

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,108 @@
...
+
+char& MyString::operator[](int idx)
+{
+	return mBuffer.get()[idx];
+}
+
+MyString MyString::operator+(const MyString& str) const
+{
+	if (empty()) return MyString(str);
+	else if (str.empty()) return MyString(*this);
```
karunrao97: Is it okay to just return the object directly instead of invoking the copy constructor (i.e. will the return be copied by value anyway)?

In short, can I instead do this?:
if (empty()) return str;
else if (str.empty()) return *this;

maciekgajewski: It is OK to return the object directly, it's going to be copied anyways.
It is possible that compiler will decide to elide the extra copy you wrote, but from the "abstract C machine" indeed, there are two copy constructions taking place.

PR #24, opened 2016-05-18
MyString 2.0: homework 3

dancollier:


course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,105 @@
+#include <iostream>
+#include <cstring>
+#include "my_string.h"
+
+MyString::MyString(){
+	buffer = std::unique_ptr<char[]>(new char[1]);
```
maciekgajewski, line 6: This is correct, but it would be "nicer" if the default constructor wouldn't perform any allocations.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,105 @@
...
+	buffer[0] = '\0';
+}
+
+MyString::MyString(const char* theString){
+	buffer = std::unique_ptr<char[]>(new char[std::strlen(theString) + 1]);
+	unsigned int i;
+	for (i = 0; i < std::strlen(theString); i++){
```
maciekgajewski, line 13: wold be slightly simpler if you'd calculated the length once and the re-use it 3 times

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,105 @@
...
+	buffer = std::unique_ptr<char[]>(new char[theString.size() + 1]);
+	unsigned int i;
+	for (i = 0; i < std::strlen(theString.buffer.get()); i++){
+		buffer[i] = theString[i];
+	}
+	buffer[i] = '\0';
+}
+   
+const bool MyString::empty() const {
+	if (buffer){
```
maciekgajewski, line 29: can buffer ever be null in your string? It looks like whichever path you take, it should never be null, so the check is not needed.
You could replace this check with assert, then it will only be performed in debug mode, and in production code it will disappear.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,105 @@
...
+		std::strcpy(buffer.get(), theString.buffer.get());
+	}
+	return *this;
+}
+
+
+MyString MyString::operator + (const MyString& theString) const {
+	MyString newString; 
+	if (theString.empty()){
+		std::strcpy(newString.buffer.get(), buffer.get());
```
maciekgajewski, line 88: Can you ass a unit test for this branch? It would look like this:

```
MyString x = "qqq";
MyString y = x + MyString();
assert(y == "qqq");
```

See if it works, see if valgrind doesn't give you any memory problems.

dancollier, line 88: On Linux x86_64, it passed the tests but valgrind had errors. On MacOS x86_64, it had a segmentation fault.

The last class provided a solution which cleared all the tests, to be submitted with next homework.

PR #25, opened 2016-05-18
Homework string Alex

neuraltoys:


neuraltoys:
the only test that fails is the +=
I saw you implemented it by creating a new buffer, putting the string in and then swapping
I tried to implement it by replacing the current buffer and something didn't work, not sure why
I'm going to ask about it tomorrow

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,30 @@
...
+	MyString(const std::size_t size);
+	MyString(const MyString&);
+	
+	std::size_t size() const;
+	bool empty() const;
+	const char* c_str() const;
+	
+	bool operator==(const MyString& other) const;
+	bool operator!=(const MyString& other) const;
+	char& operator[](const std::size_t i) const;
```
maciekgajewski: This function is not const-correct. It is 'const', so it promises the uses that it will ot modify the string, yet it returns non-const reference to the strings content, which can be used to modify individual characters.

There should be two version of the operator, const and non-const.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,101 @@
...
+#include <stdexcept>
+
+MyString::MyString(): mSize(0) { }
+
+MyString::MyString(const char *inchar) :mSize(std::strlen(inchar)), mBuffer(new char[mSize + 1])
+{
+	std::strcpy(mBuffer.get(), inchar);
+}
+
+MyString::MyString(const std::size_t size) :mSize(size), mBuffer(new char[mSize + 1])
```
maciekgajewski: This is unusual.
1. What is the content of the string after this constructor is used?
2. It allows for the following code:

```
MyString sss = 5;
// what is in sss now???
```

and also for this:

```
void someFunction(const MyString& s);

someFunction(42);
```

maciekgajewski: If you find this constructor useful n your code, maybe you could make it private, so it isn't part of the class' API

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,101 @@
...
+	std::strcpy(str.mBuffer.get(), this->mBuffer.get());
+	std::strcpy(str.mBuffer.get() + this->mSize, other.mBuffer.get());
+	return str;
+}
+
+MyString& MyString::operator+=(const MyString &other) {
+	auto current_size = this->mSize;
+	this->mSize = current_size + other.mSize;
+	auto current = this->c_str();
+	this->mBuffer.reset(new char[this->mSize + 1]);
```
maciekgajewski: What does 'current' points to?
What happens after you call 'reset' on mBuffer?

neuraltoys:
I've fixed the code based on your comments and added the begin and end implementation

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,33 @@
...
+    void operator=(const char* inchar);
+    MyString& operator=(const MyString& other);
+    MyString operator+(const MyString& other) const;
+    MyString& operator+=(const MyString& other);
+
+    char* begin();
+    char* end();
+
+private:
+    std::size_t mSize;
```
maciekgajewski, line 31: If you decided to use `this->` everywhere, there is no need to use the 'm' prefix

PR #27, opened 2016-05-22
Willis MyString

wiggity:
Added move assignment operator, move constructor, and more comparison operators.


wiggity:
should i assume that with no comments my pull request is perfect? or more likely this slipped through the cracks.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,129 @@
...
+
+bool MyString::operator>=(const MyString& rhs) const{
+	if (!buf)
+		return false;
+	if (!rhs.buf)
+		return true;
+	return strcmp(buf.get(), rhs.buf.get()) >= 0;
+}
+
+bool MyString::operator<(const MyString& rhs) const{
```
maciekgajewski, line 76: You only need "==" and "<", all the others a) can be implemented in terms of !=, b) are much less needed

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,129 @@
+#include "my_string.h"
+#include <cstring>
+#include <stdexcept>
+
+MyString::MyString(const char* str) : buf(std::make_unique<char[]>(std::strlen(str)+1)) {
+	std::strcpy(buf.get(), str);
```
maciekgajewski, line 6: Would be good to check if str is not empty

PR #28, opened 2016-05-23
Homework Duncan 6

dgkimpton:
String homework from meeting 6


course3/homework-string/build.bat
```diff
@@ -0,0 +1,22 @@
+@echo off
```
maciekgajewski, line 1: you should use cmake instead

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,192 @@
+#include "my_string.h"
+
+#include <cstring>
+#include <memory>
+
+MyString::MyString() 
+	:
+	mBuffer(nullptr)
```
maciekgajewski: this is not necessary, as any unique_ptr is null by default (initialized to null in it's own default constructor).
This means that the entire constructor is unnecessary.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,192 @@
...
+MyString::MyString(MyString&& that)
+{
+	mBuffer.swap(that.mBuffer);
+}
+
+MyString::MyString(const char* str)
+{
+	if (str == nullptr || str[0] == '\0')
+	{
+		mBuffer = nullptr;
```
maciekgajewski: is this required?

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,192 @@
...
+	else
+	{
+		mBuffer = std::make_unique<char[]>(std::strlen(str) + 1);
+		std::strcpy(mBuffer.get(), str);
+	}
+}
+
+MyString& MyString::operator=(const MyString& that)
+{
+	if (this == &that)
```
maciekgajewski: interesting test. Why do you think it is needed?

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,192 @@
...
+		return *this;
+	
+	if (that.empty())
+	{
+		std::unique_ptr<char[]> oldBuffer = nullptr;
```
maciekgajewski: there is no need to initialize unique_ptr to null

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,192 @@
...
+		mBuffer.swap(oldBuffer);
```
maciekgajewski: these two lines are a very elaborate way of saying

```
mBuffer.reset();
```

Here's some doco: http://en.cppreference.com/w/

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,192 @@
...
+	}
+	else
+	{
+		auto newBuffer = std::make_unique<char[]>(std::strlen(that.mBuffer.get()) + 1);
+		std::strcpy(newBuffer.get(), that.mBuffer.get());
+
+		mBuffer.swap(newBuffer);
```
maciekgajewski: Technically correct, but the code would be cleaner if you'd say

```
mBuffer = std::move(newBuffer);
```

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	return *this;
+}
+
+MyString& MyString::operator=(MyString&& that)
+{
+	if (this == &that)
```
maciekgajewski: Where coud this possibly occur? Why do you need to check for this? What would happen if you wouldn't?

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,192 @@
...
+	auto thisSize = size();
+	auto thatSize = that.size();
+	
+	if (thisSize == 0 && thatSize == 0)
+	{
+		return MyString();
+	}
+	else if (thatSize == 0)
+	{
+		return MyString(mBuffer.get());
```
maciekgajewski: Why `MyString(mBuffer.get())` instead of `*this`?
Why `MyString(that.mBuffer.get())` instead of `that`?

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,192 @@
...
+	}
+	else if (thisSize == 0)
+	{
+		return MyString(that.mBuffer.get());
+	}
+	else
+	{
+		auto result = MyString();
```
maciekgajewski: `MyString result;`

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,192 @@
...
+}
+
+char& MyString::operator[](int index)
+{
+	return mBuffer[index];
+}
+
+const char* MyString::c_str() const
+{
+	return mBuffer.get();
```
maciekgajewski: And what will happen if buffer is null?

course3/homework-vector/vector.cc
```diff
@@ -3,9 +3,59 @@
 
 class Vector
 {
-    // TODO
```
maciekgajewski, line 6: please, sent assignemtnsd in separate pul requests. keep them in separate branches

maciekgajewski:
what happened to all my comments?

dgkimpton:
I don't know where they went, but I did get an e-mail with each of them so
you definitely entered them...

On 23 May 2016 at 17:39, maciekgajewski notifications@github.com wrote:

> what happened to all my comments?
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> https://github.com/maciekgajewski/c-excercises/pull/28#issuecomment-221013331

## 

Mr Duncan Kimpton
e-mail: dgkimpton@gmail.com

course3/homework-string/build.bat
dgkimpton, line 1: I started out that way, but I couldn't get it to work without needing a .bat to run cmake correctly... in the end it was getting overly complex for something very simple. So for now I went with a simple .bat, to be revisited when the projects get more complex.

I had something like this:

```
project(test)

cmake_minimum_required(VERSION 3.5.2)
set(CMAKE_CXX_COMPILER C:/MinGW/bin/g++.exe)
set(CMAKE_C_COMPILER C:/MinGW/bin/gcc.exe)
set(CMAKE_LINKER C:/MinGW/bin/ld.exe)
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall -Werror -pedantic -g")


add_executable(${PROJECT_NAME}
test.cpp
)
```

But then I had to run

```
cmake ../ -G "MSYS Makefiles"
```

to get it to build the make file before running make, which I kept messing up.

course3/homework-string/my_string.cc
dgkimpton: OK, clear.

course3/homework-string/my_string.cc
dgkimpton: The if yes, the assignment not.

course3/homework-string/my_string.cc
dgkimpton: I thought it was needed in case I did something like:

```
MyString a("Test");
MyString& x = a;
a = x;
```

course3/homework-string/my_string.cc
dgkimpton: Indeed, this one repeats a fair bit, sorry.

course3/homework-string/my_string.cc
dgkimpton: cool :)

course3/homework-string/my_string.cc
dgkimpton: wouldn't the consumer expect a copy to happen? Otherwise later operations on the result could modify the input.

y = a+b;
y[3] = 't';

course3/homework-string/my_string.cc
dgkimpton: indeed.

course3/homework-string/my_string.cc
dgkimpton: gah, indeed.

course3/homework-vector/vector.cc
dgkimpton, line 6: Huh, I think this must be a hang over from when I started doing all the assignments in master... those are now in the new branch I created. I'll see what I can do about that for the next time.

course3/homework-string/my_string.cc
dgkimpton: I'm not sure, I copied it over from my normal operator= method. And indeed if I didn't it would just swap the pointer with itself and do nothing. So really not needed.

course3/homework-string/build.bat
maciekgajewski, line 1: You shouldn't set compiler in CMake, that CMake's job to figure that out.

You need to run cmake to generate makefile (or ninja file, Visual Studio solution etc etc), but usually you do it only once.

On Windows - where things tend to get unnecessary complex - you may have to use correct "prompt" - a shell instance with some environment variables set. MinGW should come with one.

Using cmake should be a simple thing.

course3/homework-string/my_string.cc
maciekgajewski: Think of the "don't pay for what you are not using" principle. How frequently would this mistake happen (very rarely) vs how often do you need to pay the price of the check (always)

I think it would be nice if your function simply worked correctly in such a case, but it's fine if it doesn't

course3/homework-string/my_string.cc
maciekgajewski: But this function returns MyString by copy, so copy will happen.
Remember - every type here is a "value type", every return value and parameter is passed "by value"

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,181 @@
...
+
+char& MyString::operator[](int index)
+{
+	return mBuffer[index];
+}
+
+const char* MyString::c_str() const
+{
+	if (mBuffer == nullptr)
+		return '\0';
```
maciekgajewski: This won't work. What is the return type? What is the type of `'\0'`? What do you get when you convert the value from one (integer) type to another (integer) type?

course3/homework-string/my_string.cc
dgkimpton: Interesting, I see your point. Makes sense.

course3/homework-string/my_string.cc
dgkimpton: Indeed it does, your way is clearly superiour with that in mind.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,192 @@
...
+	if (thisEmpty && thatEmpty)
+		return false;
+
+	if (thisEmpty && !thatEmpty)
+		return true;
+
+	if (!thisEmpty && thatEmpty)
+		return false;
+	
+	return std::strcmp(mBuffer.get(), that.mBuffer.get()) < 1;
```
maciekgajewski: Once you ensure c_str() works correctly (always returns valid pointer), you cna reduce the entire function to this single line:

```
return std::strcmp(c_str(), that.c_str()) < 0
```

maciekgajewski: Why <1? Shouldn't it be <0 ?

PR #29, opened 2016-05-26
unique_ptr

karunrao97:


course3/homework-unique_ptr/test.cc
```diff
@@ -0,0 +1,99 @@
+#include "unique_ptr.h"
+#include "unique_ptr.cc"
+
+#include <cassert>
+#include <iostream>
+#include <memory>
+
+
+template<template<class...> class UniquePtrType>
```
karunrao97: I arrived at this via brute force by trying a bunch of different template definitions until my code compiled. I don't quite understand exactly what's going on here, especially the ellipsis (...). If I drop the ellipsis, the code does not compile, but if I also comment out line 92 (unique_ptr_test<std::unique_ptr>();), the code does compile and passes all tests, so it seems like it's only required for std::unique_ptr, and not for karun::unique_ptr.

I also couldn't figure out how to pass a generic make_unique function as a template parameter (my goal was to change line 92 to unique_ptr_test<std::unique_ptr, std::make_unique>();). I think I need a template template function here, but it looks like that is not a legal template parameter. Is there another approach I could take here? I tried many variants of the following - it doesn't compile, and hurts my brain every time I look at it:
template<template<class T> class UniquePtrType, UniquePtrType<T> (*MakeUnique)(T*)>

maciekgajewski: True; template-templates are fancy and complex. You need to have good reason to use them. In this case a valid strategy would be to a) not make this a template function, simply test your pointer only, b) assume fixed pointed-to type:

```
template<typename PtrType>
void unique_ptr_test()
{
// ...
}

// ...
unique_ptr_test<std::unique_ptr<int>>();
unique_ptr_test<karun::unique_ptr<int>>();
```

course3/homework-unique_ptr/test.cc
```diff
@@ -0,0 +1,99 @@
...
+	assert(ptr.get() == nullptr);
+	
+	// initialization constructor
+	UniquePtrType<int> ptr2(nullptr);
+	assert(!ptr2);
+	
+	int* i = new int(42);
+	UniquePtrType<int> ptr3(i);
+	assert(ptr3);
+	assert(*ptr3.get() == 42);
```
maciekgajewski: get shouldn't be required. a smart pointer should mimic the syntax of raw pointer. so `assert(*ptr3 == 42);`

course3/homework-unique_ptr/test.cc
```diff
@@ -0,0 +1,99 @@
...
+	
+	// operator*
+	struct S { int i = 42; };
+	UniquePtrType<S> ptr5(new S);
+	S s = *ptr5;
+	assert(s.i == 42);
+	
+	// operator->
+	assert(ptr5->i == 42);
+
```
maciekgajewski: Pretty good!
One important features is not tested thou: if the pointed object is destroyed when the unique_ptr is destroyed, assigned-to or `reset` is called.

course3/homework-unique_ptr/test.cc
```diff
@@ -0,0 +1,99 @@
+#include "unique_ptr.h"
+#include "unique_ptr.cc"
```
maciekgajewski: This is why template classes don't have the .cc file, and are either entirely implementd in the .h. or - like in - your case, in the header files: one containing the class definition, and the other method definitions.

Different naming convetiosn are used for the second file: XXX_impl.h, XXX.hxx, XXX.ipp

course3/homework-unique_ptr/unique_ptr.cc
```diff
@@ -0,0 +1,111 @@
+#include "unique_ptr.h"
+
+#include <utility>
+
+
+template<class T>
+karun::unique_ptr<T>::unique_ptr() : mPtr(nullptr) {}
```
maciekgajewski: You could have used `= default` in the class definition.

maciekgajewski:
Nice!

course3/homework-unique_ptr/test.cc
karunrao97: The fixed pointed-to type was my first approach, and my code was exactly as you suggested above. However, in order to test operator->, I wanted to use a different type (struct, in my case), which is why I went for the template-template.

course3/homework-unique_ptr/test.cc
karunrao97: What's the standard naming convention used at Optiver?

course3/homework-unique_ptr/test.cc
karunrao97: I didn't implement operator\* when I wrote this test, hence the get(). `assert(*ptr3 == 42);` does work, though - I just committed a similar test in line 78.

course3/homework-unique_ptr/test.cc
karunrao97: Good point. I wanted to test this, but didn't know how. As far as I could tell from searching on Google, there is no way to check if memory has been freed. Even if the object has been destroyed, it might stay in memory and get cleaned up or reallocated at a later stage. Or is that incorrect?

I guess one way to test this might be to write a simple class that sets some global flag in its destructor. Is there a better approach?

course3/homework-unique_ptr/unique_ptr.cc
karunrao97: I did that in my first implementation. However, if I use default, I also need to initialize mPtr to nullptr in the header file, else it contains a junk pointer. Since most other constructors overwrite the initial value in mPtr anyway, I decided to eliminate the initialization in the header file and just ensure all constructors initialize mPtr appropriately. This is probably a completely unnecessary optimization - is there actually any runtime cost to initializing a member in the header file and then overwriting it in a constructor?

course3/homework-unique_ptr/unique_ptr.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace karun
+{
+	template <class T>
+	class unique_ptr
+	{	
+		
+	public:
+		unique_ptr();
+		unique_ptr(std::nullptr_t);
```
karunrao97: I forgot to ask this before, but even if I remove the `unique_ptr(std::nullptr_t)` constructor, calling `unique_ptr<int> p(nullptr);` creates a new unique_ptr p with p.mPtr equal to nullptr. I was surprised by this, and expected the code to not compile since the only compatible constructor (`unique_ptr(T*)`) is marked as explicit - is nullptr a special case, able to implicitly cast to T\* despite the explicit specifier?

course3/homework-unique_ptr/test.cc
maciekgajewski: XX_impl.h

course3/homework-unique_ptr/test.cc
maciekgajewski: You don't really need to check whether the memory was freed, you need t check if 'delete' has been called on the object, because calling 'delete' is the only responsibility of the programmer who used 'new'.

How I usually do it, I have a small class that has a counter as a static member variable, and I increase the counter in constructors (default and copy), and decrease it in destructor. This way you can check how many instances currently exists.

course3/homework-unique_ptr/unique_ptr.cc
maciekgajewski: There shouldn't be any. And it's fine to set the value to nullptr in class definition. The value in class definition is the one used during default construction of the member; in you other constructors you perform non-default initialization.

course3/homework-unique_ptr/unique_ptr.h
maciekgajewski: http://en.cppreference.com/w/cpp/language/nullptr

> There exist implicit conversions from nullptr to null pointer value of any pointer type

I don't think the overload for nullptr_t is required at all.

course3/homework-unique_ptr/test.cc
karunrao97: Nice! Will do.

course3/homework-unique_ptr/unique_ptr.h
karunrao97: Shouldn't the explicit specifier prevent this implicit conversion from nullptr_t to T*, though? Or perhaps I misunderstand how explicit works.

course3/homework-unique_ptr/unique_ptr.cc
karunrao97: Thanks, will change this.

course3/homework-unique_ptr/unique_ptr.h
maciekgajewski: The 'explicit' specifier prevent implicit conversion from another type into your type. In this case the conversion is from nullptr_t to pointer.

karunrao97: Ah okay. I thought it prevented all implicit conversions, which doesn't make sense in retrospect. It's clear now, thanks!

PR #30, opened 2016-05-29
Homework cc

cc-catch:
test & initial scratch

### DIFF
--- a/course3/homework-string/CMakeLists.txt
+++ b/course3/homework-string/CMakeLists.txt
@@ -1,0 +1,9 @@
+project(MyString)
+
+cmake_minimum_required(VERSION 2.8)
+
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -Wall -std=c++14")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -g -std=c++14")
+
+set(SOURCES test.cc my_string.cpp my_string.h)
+add_executable(stringtest ${SOURCES})
--- a/course3/homework-string/my_string.cpp
+++ b/course3/homework-string/my_string.cpp
@@ -1,0 +1,175 @@
+#include "my_string.h"
+#include <memory>
+
+        MyString::MyString()
+        {
+            mBuffer=nullptr;
+        }
+
+        MyString::MyString(const char* other)
+        {
+            mBuffer = std::make_unique<char[]>(std::strlen(other) + 1);
+            std::strcpy(mBuffer.get(), other);
+        }
+
+        MyString::MyString(const MyString& other)
+        {
+            mBuffer = std::make_unique<char[]>(other.size() + 1);
+            std::strcpy(mBuffer.get(), other.mBuffer.get());
+        }
+
+        std::size_t MyString::size() const
+        {
+            if(!mBuffer)
+                return 0;
+
+            return std::strlen(mBuffer.get());
+        }
+
+        bool MyString::empty() const
+        {
+            return mBuffer == nullptr || mBuffer.get()[0] == 0;
+        }
+
+        char MyString::operator[](std::size_t index) const
+        {
+            return mBuffer.get()[index];
+        }
+        char& MyString::operator[](std::size_t index)
+        {
+            return mBuffer.get()[index];
+        }
+
+        bool MyString::operator==(const MyString& other) const
+        {
+            if(this->empty() && other.empty())
+                //TODO: this is not completely right just yet. we need nullptr checks for later on.
+                return true;
+
+            return std::strcmp(mBuffer.get(), other.mBuffer.get()) == 0;
+        }
+        bool MyString::operator!=(const MyString& other) const
+        {
+            return !(*this==other);
+        }
+
+        MyString& MyString::operator=(const MyString& other)
+        {
+            if(!other.mBuffer)
+            {
+                clearBuffer();
+            }
+            else
+            {
+                auto newSize = other.size() + 1;
+                auto newBuffer = std::make_unique<char[]>(newSize);
+                std::strcpy(newBuffer.get(), other.mBuffer.get());
+                mBuffer.swap(newBuffer);
+            }
+
+            return *this;
+        }
+
+        MyString& MyString::operator=(const char* other)
+        {
+            if(other == nullptr)
+            {
+                clearBuffer();
+            }
+            else
+            {
+                auto newSize =std::strlen(other) + 1;
+                auto newBuffer = std::make_unique<char[]>(newSize);
+
+                std::strcpy(newBuffer.get(), other);
+                mBuffer.swap(newBuffer);
+            }
+
+            return *this;
+        }
+
+        const char* MyString::c_str() const
+        {
+            if(!mBuffer)
+                return nullptr;
+
+            return mBuffer.get();
+        }
+
+        MyString& MyString::operator+=(const char* other)
+        {
+            if(other != nullptr && other[0] != '\0')
+            {
+                auto newSize = this->size() + std::strlen(other) + 1;
+                auto copyStartIndex = this->size();
+
+                auto newBuffer = std::make_unique<char[]>(newSize);
+                std::strcpy(newBuffer.get(), this->c_str());
+                std::strcpy(newBuffer.get() + copyStartIndex, other);
+
+                mBuffer.swap(newBuffer);
+            }
+
+            return *this;
+        }
+
+        MyString& MyString::operator+=(const MyString& other)
+        {
+            if(!other.empty())
+            {
+                auto newSize = this->size() + other.size() + 1;
+                auto copyStartIndex = this->size();
+
+                auto newBuffer = std::make_unique<char[]>(newSize);
+                std::strcpy(newBuffer.get(),this->c_str());
+                std::strcpy(newBuffer.get() + copyStartIndex, other.c_str());
+                mBuffer.swap(newBuffer);
+            }
+
+            return *this;
+        }
+
+        MyString MyString::operator+(const char* other) const
+        {
+            if(other == nullptr || other[0] == 0)
+            {
+                return MyString(*this);
+            }
+            else
+            {
+                auto newSize = this->size() + std::strlen(other) + 1;
+                auto copyStartIndex = this->size();
+                auto newBuffer = std::make_unique<char[]>(newSize);
+                std::strcpy(newBuffer.get(), mBuffer.get());
+                std::strcpy(newBuffer.get() + copyStartIndex, other);
+
+                auto result = MyString(newBuffer.get());
+                return result;
+            }
+        }
+
+        MyString MyString::operator+(const MyString& other) const
+        {
+            if(other.empty())
+            {
+                return MyString(*this);
+            }
+            else
+            {
+                auto newSize = this->size() + other.size() + 1;
+                auto copyStartIndex = this->size();
+
+                auto newBuffer = std::make_unique<char[]>(newSize);
+                std::strcpy(newBuffer.get(), mBuffer.get());
+                std::strcpy(newBuffer.get() + copyStartIndex, other.mBuffer.get());
+
+                return MyString(newBuffer.get());
+            }
+        }
+
+
+        void MyString::clearBuffer()
+        {
+            //TODO: what happens with the previous pointer?
+            mBuffer = nullptr;
+        }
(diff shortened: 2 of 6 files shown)


course3/homework-string/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+project(MyString)
+
+cmake_minimum_required(VERSION 2.8)
+
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -Wall -std=c++14")
```
maciekgajewski, line 5: Don't re-define CXX_FLAGS_DEBU, unless you are doing something fancy

course3/homework-string/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
...
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -g -std=c++14")
```
maciekgajewski, line 6: Don't add '-g' to the default set of flags. CMake will add it by itself when you configure the project in debug mode.

course3/homework-string/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+project(MyString)
+
+cmake_minimum_required(VERSION 2.8)
```
maciekgajewski, line 3: 2.8 is ancient, go for 3.0

course3/homework-string/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
...
+
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -Wall -std=c++14")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -g -std=c++14")
+
+set(SOURCES test.cc my_string.cpp my_string.h)
+add_executable(stringtest ${SOURCES})
```
maciekgajewski, line 9: Why the variable? Doing it all with one command would be cleaner.

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
+#include "my_string.h"
+#include <memory>
+
+        MyString::MyString()
```
maciekgajewski, line 4: Why indenting the entire file? This doesn't improve readability, only wastes significant portion of precious screen space.

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,39 @@
-#pragma once
+#ifndef MY_STRING_H
```
maciekgajewski, line 1: What's wrong with `#pragma once`?

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+        {
+            mBuffer=nullptr;
```
maciekgajewski, line 6: This doesn't do anything. The smart pointer is initialized to null if default-constructed.

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+        MyString::MyString(const char* other)
+        {
+            mBuffer = std::make_unique<char[]>(std::strlen(other) + 1);
+            std::strcpy(mBuffer.get(), other);
+        }
+
+        MyString::MyString(const MyString& other)
+        {
+            mBuffer = std::make_unique<char[]>(other.size() + 1);
+            std::strcpy(mBuffer.get(), other.mBuffer.get());
```
maciekgajewski, line 18: And what if `other.mBuffer` is null?

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+        MyString::MyString(const char* other)
+        {
+            mBuffer = std::make_unique<char[]>(std::strlen(other) + 1);
+            std::strcpy(mBuffer.get(), other);
+        }
+
+        MyString::MyString(const MyString& other)
+        {
+            mBuffer = std::make_unique<char[]>(other.size() + 1);
```
maciekgajewski, line 17: Does it make sense if `other` is empty?

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+        {
+            if(!mBuffer)
+                return 0;
+
+            return std::strlen(mBuffer.get());
+        }
+
+        bool MyString::empty() const
+        {
+            return mBuffer == nullptr || mBuffer.get()[0] == 0;
```
maciekgajewski, line 31: or just `mBuffer[0]`

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+            return mBuffer.get()[index];
+        }
+        char& MyString::operator[](std::size_t index)
+        {
+            return mBuffer.get()[index];
+        }
+
+        bool MyString::operator==(const MyString& other) const
+        {
+            if(this -> empty() && other.empty())
```
maciekgajewski, line 53: Style: this time you use `this->`, and in other methods you don't. Make your choice and stick to it.

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+                //TODO: this is not completely right just yet. we need nullptr checks for later on.
```
maciekgajewski, line 54: Why? Which pointers can be null?

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+            return std::strcmp(mBuffer.get(), other.mBuffer.get()) == 0;
+        }
+        bool MyString::operator!=(const MyString& other) const
+        {
+            return !(*this==other);
+        }
+
+        bool MyString::operator<(const MyString& other) const
+        {
+            if(this->empty() && other.empty())
```
maciekgajewski, line 66: is this test needed?

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+                return true;
+
+            return std::strcmp(this->c_str(), other.c_str()) <= 0;
+        }
+
+        bool MyString::operator>(const MyString& other) const
+        {
+            if(this->empty() && other.empty())
```
maciekgajewski, line 82: There's no point implementing all the other operators, `<` and `==` are all you ever need, the rest you _could_ implement in terms of the two, but you wouldn't have to, because you could use Boost.Operators to do that for you.

http://www.boost.org/doc/libs/1_59_0/libs/utility/operators.htm

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+                std::strcpy(newBuffer.get(), other.mBuffer.get());
+                mBuffer.swap(newBuffer);
+            }
+
+            return *this;
+        }
+
+        MyString& MyString::operator=(const char* other)
+        {
+            if(other == nullptr)
```
maciekgajewski, line 115: I think you could require to user to never provide null pointer as a param, and omit the check. Remember the "dont pay for what you are not using" principle

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+            {
+                clearBuffer();
+            }
+            else
+            {
+                auto newSize =std::strlen(other) + 1;
+                auto newBuffer = std::make_unique<char[]>(newSize);
+
+                std::strcpy(newBuffer.get(), other);
+                mBuffer.swap(newBuffer);
```
maciekgajewski, line 125: `mBuffer = std::move(newBuffer)` would work in the same way, and the code would be cleaner to read.

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+            }
+
+            return *this;
+        }
+
+        MyString& MyString::operator+=(const char* other)
+        {
+            if(other != nullptr && other[0] != '\0')
```
maciekgajewski, line 133: isn't this the same as `if(!other.empty())` ?

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+            {
+                auto newSize = this->size() + other.size() + 1;
+                auto copyStartIndex = this->size();
+
+                auto newBuffer = std::make_unique<char[]>(newSize);
+                std::strcpy(newBuffer.get(), mBuffer.get());
+                std::strcpy(newBuffer.get() + copyStartIndex, other.mBuffer.get());
+
+                return MyString(newBuffer.get());
+            }
```
maciekgajewski, line 199: Lots of repeated code in the operators above. Couldn't it be factored out, or maybe one operator could be implemented in terms of another.

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+
+        void MyString::clearBuffer()
+        {
+            //TODO: what happens with the previous pointer?
```
maciekgajewski, line 205: What "previous pointer" do you have in mind?

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+            mBuffer = nullptr;
```
maciekgajewski, line 206: This is the same as `mBuffer.reset()`

course3/homework-string/my_string.cpp
```diff
@@ -0,0 +1,209 @@
...
+        }
```
maciekgajewski, line 207: Is there a point defining separate method just for this one operation?

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,39 @@
...
+#define MY_STRING_H
+#include<iostream>
```
maciekgajewski, line 3: What are you using iostream for?

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,39 @@
...
+#include<memory>
+#include<cstring>
```
maciekgajewski, line 5: You don't use cstring in this file

course3/homework-string/my_string.cpp
cc-catch, line 4: Fair. I must've messed something in qt creator.

course3/homework-string/my_string.h
cc-catch, line 1: Nothing actually. The remnant of creating the project as new in qt creator I guess.

course3/homework-string/my_string.cpp
cc-catch, line 6: Correct. Need to get used to the more non-verbose c++ style.

course3/homework-string/my_string.cpp
cc-catch, line 18: Then it's a bug :(

course3/homework-string/my_string.cpp
cc-catch, line 17: Best to reset it then, indeed. This works, but you're right.

course3/homework-string/my_string.cpp
cc-catch, line 54: I meant more like checking of the empty-ness is different

course3/homework-string/my_string.cpp
cc-catch, line 133: other is a const char\* here.

course3/homework-string/my_string.cpp
cc-catch, line 199: Yes. Tidying up stage coming. I still have some TODO's left as well

course3/homework-string/my_string.cpp
cc-catch, line 205: I meant the previous unique ptr. The comment is obsolete now - the previous value, if any, will go out of scope so it gets cleaned properly.

course3/homework-string/my_string.cpp
cc-catch, line 206: That's one thing I was missing :) I was trying to do all of this out of memory and now I'm revisiting the notes.

course3/homework-string/my_string.cpp
cc-catch, line 207: Not anymore. See my previous comment - I was trying to write the homework without the notes, and I couldn't remember how to clear the unique ptr. So I encapsulated it in a method until a later moment.

course3/homework-string/my_string.h
cc-catch, line 3: I used to have some std::cout for testing

course3/homework-string/my_string.cpp
maciekgajewski, line 133: ah yes, my bad

course3/homework-string/my_string.cpp
maciekgajewski, line 206: http://en.cppreference.com/w/

PR #32, opened 2016-06-01
Move const mau

zerokill:


course3/homework-string/my_string.h
```diff
@@ -1,6 +1,110 @@
...
+            return "";
+    }
+
+    MyString& operator=(const MyString& rhs) {
+        if (rhs.empty()) {
+            buff.reset();
+        }
+        else {
+            buff = std::make_unique<char []>(rhs.length + 1);
+            std::strcpy(buff.get(), rhs.buff.get());
```
maciekgajewski, line 62: What about length? Punitive excercise: write a unit test that will fail because of this bug, and only then fix it.

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,110 @@
...
+        }
+        return *this;
+    }
+
+    MyString& operator=(MyString&& rhs) {
+        buff = std::move(rhs.buff);
```
maciekgajewski, line 69: again: missing length. See my previous commet

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,110 @@
...
+
+    MyString& operator+=(const MyString& rhs) {
+        if (!rhs.empty()) {
+            std::size_t currentSize = size();
+            std::unique_ptr<char []> temp = std::make_unique<char []>(currentSize + rhs.size() + 1);
+            if (size())
+                std::strcpy(temp.get(), buff.get());
+            if (rhs.size())
+                std::strcpy(temp.get() + currentSize, rhs.buff.get());
+            buff.swap(temp);
```
maciekgajewski, line 88: This would work, but `buff = std::move(temp);` looks more expressive, reads more naturally.

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,110 @@
...
+            length += rhs.size();
+        }
+
+        return *this;
+    }
+
+    bool operator<(const MyString& rhs) {
+        if (!buff)
+            return true;
```
maciekgajewski, line 97: won't work if the other guy is empty as well

course3/homework-string/my_string.h
```diff
@@ -1,6 +1,110 @@
...
+            return std::strcmp(buff.get(), rhs.buff.get()) == 0;
+        // if this or rhs is null this will return false
+        return buff.get() == rhs.buff.get();
+    }
+
+    char& operator[](int index) {
+        return buff[index];
+    }
+
+    char& operator[](int index) const {
```
maciekgajewski, line 118: this shouldn't return mutable reference, but const ref or copy. Otherwise it would be possible to modify const object, which would be a nasty surprise for your user.

course3/homework-string/test.cc
```diff
@@ -26,6 +26,9 @@ void string_test()
 	
 	assert(string2 == string3);
 	assert(string3.size() == 3);
+
+    assert(string2 != string1);
+    assert(string1 != string2);
```
maciekgajewski, line 33: your indentation is wrong; doesn't match the rest of the file.

course3/homework-vector/vector.cc
```diff
@@ -1,11 +1,45 @@
...
 #include <cassert>
+#include <cmath>
 
 class Vector
 {
-    // TODO
+    int mX, mY;
+
+public:
+    Vector() {}
```
maciekgajewski, line 10: This fails to initialize the members, leaving them unspecified.

course3/homework-vector/vector.cc
```diff
@@ -1,11 +1,45 @@
...
+    Vector operator+(const Vector& rhs) {
+        int sumX = this->GetX() + rhs.GetX();
+        int sumY = this->GetY() + rhs.GetY();
+
+        return Vector(sumX, sumY);
+    }
 };
 
+std::ostream& operator<<(std::ostream& out, const Vector& vec) {
+    return out << "{x= " << vec.GetX() << " , y= " << vec.GetY() << "}" << std::endl;
```
maciekgajewski, line 39: Don't add `std::endl` inside of your streaming operator. What if your user wants to print multiple vectors in one line?

course3/homework-vector/vector.cc
```diff
@@ -1,11 +1,45 @@
...
+    int mX, mY;
+
+public:
+    Vector() {}
+    Vector(int x, int y) : mX(x), mY(y) {}
+
+    int GetX() const { return mX; }
+    int GetY() const { return mY; }
+
+    double GetPolarLength() const { return sqrt((mX * mX) + (mY * mY)); }
```
maciekgajewski, line 16: It's nicer to use qualified versions of the old C functions: `std::sqrt`, `std::atan2` etc.

course3/homework-string/my_string.h
zerokill, line 62: Done

course3/homework-string/my_string.h
zerokill, line 69: and done

course3/homework-string/test.cc
```diff
@@ -56,14 +56,23 @@ void string_test()
...
+    StringType stringCopy;
+
+    stringCopy = stringMau;
+    assert(stringCopy == "Maurice"); 
+    assert(stringCopy.size() == 7);
+
 	// operator +
 	StringType plusResult = string_const_ref + StringType("123");
 	assert(plusResult == "Hex123");
+    assert(plusResult.size() == 6);
```
maciekgajewski, line 77: Good! But the indentation is still wrong, it appears that you're using spaces instead of tabs. Please ,enable visual whitespace in your editor, and maintain consistency with the rest of the code.

course3/homework-string/my_string.h
```diff
@@ -37,6 +37,14 @@ class MyString
         return length;
     }
 
+    const char* begin() const {
```
maciekgajewski, line 41: a unit test for these would be useful

PR #33, opened 2016-06-01
MyString 2.1: meeting6 homework but not yet unique_ptr 

dancollier:


course3/homework-string/my_string.h
```diff
@@ -1,6 +1,44 @@
...
 	// TODO
+
+	public:
+		MyString();
+
+		MyString(const char* argString);
+
+		MyString(const MyString& argString);
+    
+        MyString(MyString&& argString);
```
maciekgajewski, line 16: inconsistent indentation

course3/homework-string/test.cc
```diff
@@ -45,6 +45,46 @@ void string_test()
...
+    StringType y = x + StringType();
+    assert(y == "qqq");
+    
+    //Test constructor MyString(const MyString& rhs){*this = otherString;}
+    StringType t1 = y; // y & t1 = qqq
+    StringType t2(t1); // t2 = &t1
+    t1 = "MyNewString";
+    assert( t2 == "qqq");
+    StringType t3;
+    if (true){
```
maciekgajewski, line 79: it's fine to simply use empty brackets to create a scope.

course3/homework-string/test.cc
```diff
@@ -53,11 +93,11 @@ int main()
 {
 	// pre-test - of this _doesn't_ crash your program, you faield to compile it in debug mode (g++ -g)
 	// if it crashes, it's a good sign. Remove the line and proceed.
-	assert(false);
+	//assert(false);
 	
 	// test fior std::string - works
 	std::cout << "Testing std::string..." << std::endl;
-	string_test<std::string>();
+	//string_test<std::string>();
```
maciekgajewski, line 100: Why? Running the same test over both string types ensures that your implementations behaves in the expected way.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,164 @@
+#include <iostream>
+#include <cstring>
+#include "my_string.h"
```
maciekgajewski, line 3: It's best to order your includes from the most specific to the most general ones. So in this case "my_string.h" should be first, and the STD headers last.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,164 @@
...
+
+MyString::MyString(){}
+
+MyString::MyString(const char* argString)
+{
+        buffer = std::unique_ptr<char[]>(new char[ std::strlen(argString) + 1]);
+        std::strcpy(buffer.get(), argString);;
```
maciekgajewski, line 10: Do you have compilation warnings enabled? Compiler should warn about the two semicolons...

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,164 @@
...
+
+
+MyString::MyString(MyString&& argString){
+        *this = std::move(argString);
+}
+
+const bool MyString::empty() const
+{
+        if (buffer){
+                return buffer[0] == '\0' || buffer[0] == ' ';
```
maciekgajewski, line 26: why do you assume that string with space as the first character should be empty?
Please ,add this to your unit test:

```
MyString nonempty = " Hey";
assert(!nonempty.empty());
```

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,164 @@
...
+        else return 0;
+}
+
+const char* MyString::c_str() const
+{
+        if (buffer){
+                return buffer.get();
+        }
+        else {
+            char* r = new char('\0');
```
maciekgajewski, line 45: And when is the buffer deleted?
As I said: any explicit "new" or "delete" in your code should be considered a bug.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,164 @@
...
+        }
+        else if ( buffer == nullptr && argString.buffer == nullptr ){
+                return true;
+        }
+        else return false;
+}
+
+const bool MyString::operator != (const MyString& argString) const
+{
+        if ( buffer && argString.buffer){
```
maciekgajewski, line 63: or you can just call operator== and negate the output.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,164 @@
...
+        else if ( buffer == nullptr && argString.buffer){
+            return true;
+        }
+        else {
+            return std::strcmp(buffer.get(), argString.buffer.get()) < 0;
+        }
+
+}
+
+char& MyString::operator[](size_t index) const
```
maciekgajewski, line 92: Do not return writable reference from const function. Otherwise you allow the user to modify the content fo const object (even by accident), which is a breach of contract.

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,164 @@
...
+{
+        MyString returnString;
+        if (buffer == nullptr && argString.buffer == nullptr){
+            return returnString;
+        }
+        else if (buffer && argString.buffer == nullptr ){
+                returnString = *this;
+        }
+        else if ( buffer == nullptr && argString.buffer){
+            returnString.buffer = std::unique_ptr<char[]>(new char[argString.size() + 1]);
```
maciekgajewski, line 133: or just say: `returnString = argString;`

course3/homework-string/my_string.cc
```diff
@@ -0,0 +1,164 @@
...
+            return *this;
+        }
+        else if (!buffer && !argString.buffer){
+            return *this;
+        }
+        else if (buffer && !argString.buffer){
+            return *this;
+        }
+        else {
+            std::strcpy(buffer.get(), argString.buffer.get());
```
maciekgajewski, line 157: Whoa! buffer is null in this branch

course3/homework-string/my_string.h
dancollier, line 16: A result of using both XCode and vi, I think. In the IDE, I don't see this issue, but I will fix it.

course3/homework-string/test.cc
dancollier, line 100: Didn't intend to leave this commented out, I was testing with std::cout output to help figure out what methods got invoked and disabled this to reduce the amount of output.

course3/homework-string/my_string.cc
dancollier, line 10: I'll have to check my XCode settings, but I'm pretty sure it was set to warn. But there were no warnings when compiling in the IDE.

course3/homework-string/my_string.cc
dancollier, line 26: Added without really thinking it through just before committing. Sleepy-minded logic error.

PR #34, opened 2016-06-04
unique_ptr and make_unique

karunrao97:


course3/homework-unique_ptr/test.cc
```diff
@@ -0,0 +1,127 @@
...
+{
+	int n;
+	static int ctr;
+	Number(int n) : n(n) { ctr++; }
+	Number(const Number& n) : n(n.n) { ctr++; }
+	~Number() { ctr--; }
+};
+int Number::ctr = 0;
+
+template<typename UniquePtr, UniquePtr (*MakeUnique)(int&&)>
```
karunrao97: How can a pass a variadic function here?
I tried `template<typename UniquePtr, UniquePtr (*MakeUnique)(int&&...)>`, which does not compile.

maciekgajewski: I don't know.. I would just simplify it

course3/homework-unique_ptr/test.cc
```diff
@@ -0,0 +1,127 @@
...
+	u2.reset(new Number(42));
+	u1.swap(u2);
+	assert(!u2);
+	assert(u1->n == 42);
+	assert(Number::ctr == 1);
+	
+	// make unique
+	u2 = MakeUnique(u1->n * 10);
+	assert(u2->n == 420);
+	assert(Number::ctr == 2);
```
maciekgajewski, line 113: You never check if Number::ctr goes back to 0 when all unique ptrs go out of scope

course3/homework-unique_ptr/unique_ptr.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+
+#include <cstddef>
+#include <utility>
+
+
+namespace karun
+{
+	template <typename T>
```
maciekgajewski: There's no point indenting the entire namepsace. It only wastes lot of screen space, and adds no readability.

course3/homework-unique_ptr/test.cc
karunrao97, line 113: I assert Number::ctr == 0 in the main function (in lines 119 and 125) immediately after the unique_ptr_test calls. Is there a way to check from within the unique_ptr_test function itself that all unique_ptr instances are correctly destroyed at the end of the unique_ptr_test scope?

course3/homework-unique_ptr/unique_ptr.h
karunrao97: Thanks, will fix

course3/homework-unique_ptr/test.cc
karunrao97: Guess I'm spoiled by Python :)

course3/homework-unique_ptr/test.cc
maciekgajewski, line 113: Ah yes, I missed that. The only way would be to enclose the entire content of the function in an additional scope, or wrap it into another function.

karunrao97, line 113: Yep, figured it out shortly after asking the question. Wrapped the content of unique_ptr_test in {}, and added an extra assert at the end of the function. I think this is nicer than my previous implementation, since all tests are contained within unique_ptr_test.

course3/homework-unique_ptr/test.cc
```diff
@@ -0,0 +1,136 @@
...
+	u1.swap(u2);
+	assert(!u2);
+	assert(u1->n == 42);
+	assert(Number::ctr == 1);
+	
+	// make_unique
+	u2 = karun::make_unique<Number>(u1->n * 10);
+	assert(u2->n == 420);
+	assert(Number::ctr == 2);
+
```
karunrao97, line 114: Added extra make_unique tests for multiple arguments (to check that the implementation is a variadic function) and non-const reference arguments (to check that they are forwarded correctly and not cast to const).

course3/homework-unique_ptr/test.cc
```diff
@@ -0,0 +1,136 @@
...
+	static int ctr;
+	Number(int n) : n(n) { ctr++; }
+	Number(int n1, int n2) : n(n1 + n2) { ctr++; }
+	Number(const Number& n) : n(n.n) { ctr++; }
+	Number(Number& n) : n(n.n * 10) { ctr++; }
+	~Number() { ctr--; }
+};
+int Number::ctr = 0;
+
+void unique_ptr_test()
```
karunrao97, line 20: Got rid of the fancy template, which makes it much easier to write tests. Would still like an easy way to quickly switch between karun::unique_ptr and std::unique_ptr when writing new tests in order to ensure their behaviour is consistent. Is it okay to do something like `using UniquePtr = karun::unique_ptr<Number>` as the first line in unique_ptr_test, so I just have to change one line if I want to try std::unique_ptr instead? If so, can I do something similar for karun::make_unique (`using MakeUnique = karun::make_unique<Number>` doesn't work since make_unique is not a type)?

maciekgajewski, line 20: Pick two types you want to point to, define two types pointers. You can use a type to group it.

```
struct std_unique_ptr_test
{
   using int_ptr = std::unique_ptr<int>;
   using number_ptr = std::unique_ptr<Number>;
   template<typename T, typename.. Args>
   std::unique_ptr<T> MakeUnique(Args&& ...args) { return std::make_unique<T>(std::forward(args)...)); }
};
struct karun_unique_ptr_test
{
   using int_ptr = karun::unique_ptr<int>;
   using number_ptr = karun::unique_ptr<Number>;
   template<typename T, typename.. Args>
   static karun::unique_ptr<T> MakeUnique(Args&& ...args) { return karun::make_unique<T>(std::forward(args)...)); }
};

template<typename Test>
void test_unique_ptr()
{
   Test::number_ptr u1;
   auto u2 = Test::make_unique<Number>(u1->n * 10);
   // ...
}

// ...
test_unique_ptr<std_unique_ptr_test>();
test_unique_ptr<karun_unique_ptr_test>();
```

karunrao97, line 20: Nice, thanks!

PR #35, opened 2016-06-05
MyString 3.0 using MyUniquePtr class

dancollier:
MyUniquePtr passes my tests (adapted from the examples for std::unique_ptr) without memory leaks, but when used with MyString, it leaks memory. Spent some time trying to fix, but not yet successfully. I think I'm ready for a hint (or the next class)...


course3/UniquePtr/Makefile
```diff
@@ -0,0 +1,17 @@
+# Makefile for MyUniquePtr and test
```
maciekgajewski, line 1: It's good that the noble art of crafting Makefiles is not lost, but I would invest some energy in mastering CMake

course3/UniquePtr/my_unique_ptr.h
```diff
@@ -0,0 +1,221 @@
...
+re-implemention of http://en.cppreference.com/w/cpp/memory/unique_ptr
+*/
+
+template<class T>
+class MyUniquePtr
+{
+public:
+    MyUniquePtr() = default;
+    
+    ~MyUniquePtr(){}
```
maciekgajewski: The sole goes of unique_ptr is to destroy the pointed-to object at the end of it's lifetime.

course3/UniquePtr/my_unique_ptr.h
```diff
@@ -0,0 +1,221 @@
...
+    
+    MyUniquePtr(T*&& ptr) : mData(ptr)
```
maciekgajewski: This makes no sense. Passing simple type (pointer) by rvalue ref doesn't really man much.
And look at the logic: you are storing a pointer to T, and immediately after you delete it. So thus constructed object holds a pointer to deleted object.

Please, write unit test, and run them via valgrind. I expect an avalanche of memory errors and leaks.

course3/UniquePtr/my_unique_ptr.h
```diff
@@ -0,0 +1,221 @@
...
+    // move
+    MyUniquePtr(T&& ptr)
+    {
+        std::move(mData, ptr.mData);
+    }
+    
+    
+    MyUniquePtr& operator=(T& other)
+    {
+        mData = other.mData;
```
maciekgajewski: What type is 'other'? What makes you think 'other' has a field called 'mData' ?

course3/UniquePtr/my_unique_ptr.h
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+
+    T *get() const
```
maciekgajewski: Please, stick the `*` to the type name, as it is a part of the type, not the method name:

```
T* get() const
```

course3/UniquePtr/my_unique_ptr.h
```diff
@@ -0,0 +1,221 @@
+#include <iostream>
+/*
+re-implemention of http://en.cppreference.com/w/cpp/memory/unique_ptr
```
maciekgajewski: Please, visit the link, and compare std::unique_ptr's interface with yours.

course3/UniquePtr/my_unique_ptr.h
```diff
@@ -0,0 +1,221 @@
...
+    {
+        return mData;
+    }
+    
+    T operator*() const
+    {
+        return *mData;
+    }
+
+    T& release()
```
maciekgajewski: release doesn't return reference to T.

course3/UniquePtr/my_unique_ptr.h
```diff
@@ -0,0 +1,221 @@
...
+    {
+        //Releases the ownership of the managed object if any. get() returns nullptr after the call.
+        auto tempData = mData;
+        mData = nullptr;
+        return *tempData;
+    }
+    
+    void reset(T& ptr)
```
maciekgajewski: Reset taks pointer, not reference to T

course3/UniquePtr/my_unique_ptr.h
```diff
@@ -0,0 +1,221 @@
...
+    }
+    
+    bool operator!=(const std::nullptr_t other) const
+    {
+        return !(*this == other);
+    }
+    
+    operator bool() const
+    {
+        return mData != nullptr ? true : false;
```
maciekgajewski: Again: the `?true:false;` is absolutely not needed.

course3/UniquePtr/my_unique_ptr.h
```diff
@@ -0,0 +1,221 @@
...
+    
+    ~MyUniquePtr()
+    {
+        //delete[] mData;
+    }
+    
+    MyUniquePtr(T*&& ptr) : mData(ptr)
+    {
+        ptr = nullptr;
+        delete[] ptr;
```
maciekgajewski: so you always call 'delete' on a null pointer?

course3/UniquePtr/my_unique_ptr.h
```diff
@@ -0,0 +1,221 @@
...
+{
+public:
+    MyUniquePtr() = default;
+    
+    ~MyUniquePtr()
+    {
+        //delete[] mData;
+    }
+    
+    MyUniquePtr(T*&& ptr) : mData(ptr)
```
maciekgajewski: This is not a moving constructor

course3/UniquePtr/ptr_test.cc
```diff
@@ -0,0 +1,73 @@
...
+    assert (std_p1 == nullptr);
+    std::array<int, 5>  std_arr = std_p1->retArray();
+    std::cout << "std_arr[1]=" << std_arr[1] << std::endl;
+    assert(std_arr[1] == 2);
+    std::cout << "std::unique_ptr test passes" << std::endl;
+    
+    std::cout << " " << std::endl;
+    
+    std::cout << "Testing MyUniquePtr..." << std::endl;
+    MyUniquePtr<Foo> my_p1(new Foo);  // p1 owns Foo
```
maciekgajewski: nope, p1 deletes Foo immediately, and keeps on holding pointer to the deleted object.

course3/UniquePtr/ptr_test.cc
```diff
@@ -0,0 +1,73 @@
...
+    MyUniquePtr<Foo> my_p2(std::move(my_p1));  // now p2 owns Foo
```
maciekgajewski: Due to the lack of move constructor on MyUniquePtr, and the fact that (implicitly defined)  copy-constructor is still available, this simply initializes my_p2 as a copy to my_p1. Something that should never happen in a correctly implemented _unique_ptr.
As the object is already deleted, and neither of the copies will attempt to perform any management on it, there will be no double delete.

course3/UniquePtr/ptr_test.cc
```diff
@@ -0,0 +1,73 @@
...
+}
+
+
+int main()
+{
+    
+    std::cout << "Testing std::unique_ptr..." << std::endl;
+    std::unique_ptr<Foo> std_p1(new Foo);  // p1 owns Foo
+    std::unique_ptr<Foo> std_p2(std::move(std_p1));  // now p2 owns Foo
+    assert(std_p1->ret6() == std_p2->ret6());
```
maciekgajewski: Here you are calling a method of a deleted object. This doesn't generate any error, because the method doesn't actually accesses any of the object's data.
Why do you expect this test to be true? std_p1 should be in a "zombie" state after being "moved from"

course3/UniquePtr/ptr_test.cc
```diff
@@ -0,0 +1,73 @@
...
+    assert( s(*std_p1) == "aString" );
```
maciekgajewski: What does this test tests? How could it fail?

course3/UniquePtr/ptr_test.cc
```diff
@@ -0,0 +1,73 @@
...
+    std_p1 = std::move(std_p2);  // ownership returns to p1
+    std::cout << "destroying std_p2...\n";
```
maciekgajewski: Why do you expect std_p2 to be destroyed?

dancollier:
Plenty of mistakes to learn from. I found the unit tests difficult for either variety of unique pointer, and knew this wasn't right because it leaked memory when used in MyString. But no memory leaks or errors in valgrind on the tests I wrote. Should've figured why before submitting. 
Thank you for the feedback!

course3/UniquePtr/ptr_test.cc
```diff
@@ -0,0 +1,73 @@
...
+#include <iostream>
+#include <cstring>
+#include <memory>
+
+
+struct Foo
+{
+    Foo()      { std::cout << "Foo::Foo\n";  }
+    ~Foo()     { std::cout << "Foo::~Foo\n"; }
+    void bar() { std::cout << "Foo::bar\n";  }
```
maciekgajewski: Neither of this methods actually touches the object's data, they all return the same constants. As such, they are useless in testing the object's identity.

maciekgajewski:
It passes the test mostly due to luck and faults in test itself. I'm not sure if the idea of unique_ptr is clear to you. If you want to dig deeper, then let's meet and talk sometime next week.

PR #38, opened 2016-06-12
list

karunrao97:


course3/homework-list/test.cc
```diff
@@ -0,0 +1,89 @@
...
+	int i = 0;
+	for (Karun::List<int>::iterator it = l.begin(); it != l.end(); ++it)
+	{
+		++i;
+		if (*it == 3) *it *= 10;
+		assert(*it == i * 10);
+	}
+	assert (i == 5);
+	
+	// const_iterator
```
karunrao97: This doesn't work, since l.begin() returns an iterator, and not a const_iterator. How do I ensure the const begin method gets called instead?

course3/homework-list/list.h
```diff
@@ -0,0 +1,106 @@
...
+		bool operator==(const iterator& it) { return mNode == it.mNode; }
+		bool operator!=(const iterator& it) { return !operator==(it); }
+		iterator& operator++() { mNode = mNode->mNext; return *this; }
+		T& operator*() { return mNode->mData; }
+		
+	private:
+		Node* mNode = nullptr;
+	};
+	
+	class const_iterator
```
karunrao97: How do I avoid this code duplication?

course3/homework-list/list.h
```diff
@@ -0,0 +1,107 @@
...
+		Node* node = new Node(data, mBegin);
+		mBegin = node;
+		mSize++;
+	}
+	
+	void pop_front()
+	{
+		Node* node = mBegin;
+		mBegin = mBegin->mNext;
+		delete node;
```
maciekgajewski: Using explicit 'delete' is a bug. Try imploementing the list without it.

course3/homework-list/list.h
```diff
@@ -0,0 +1,107 @@
...
+		Node* node = it.mNode->mNext;
+		it.mNode->mNext = it.mNode->mNext->mNext;
+		delete node;
+		mSize--;
+	}
+	
+private:
+	std::size_t mSize = 0;
+	Node* mBegin = nullptr;
+	Node* mEnd = nullptr;
```
maciekgajewski: What is the point of this variable, if tyou never assign to it?

course3/homework-list/list.h
```diff
@@ -0,0 +1,107 @@
...
+class List
+{
+public:
+	class Node
+	{
+	friend class List;
+	
+	public:
+		Node() = default;
+		Node(const Node& node) = default;
```
maciekgajewski: What is the use-case for having the node copyable?

course3/homework-list/test.cc
```diff
@@ -0,0 +1,89 @@
...
+	assert(l.front() == 20);
+	l.pop_front();
+	assert(l.size() == 1);
+	assert(l.front() == 10);
+	l.pop_front();
+	assert(l.empty());
+	assert(l.begin() == l.end());
+	
+	// front
+	l.push_front(50); l.push_front(40); l.push_front(3); l.push_front(20); l.push_front(1);
```
maciekgajewski: It would be nice being able to say:

```
l = {50, 40, 3, 20, 1};
```

course3/homework-list/test.cc
```diff
@@ -0,0 +1,89 @@
...
+	assert(l.size() == 5);
+	assert(l.front() == 1);
+	l.front() *= 10;
+	assert(l.front() == 10);
+	
+	// iterator
+	int i = 0;
+	for (Karun::List<int>::iterator it = l.begin(); it != l.end(); ++it)
```
maciekgajewski, line 48: Why don't you test the range-based for form?

```
for(int& i : l)
...
```

maciekgajewski:
Good in general, but you use explicit memory management - why? why make things harder for yourself?

course3/homework-list/list.h
karunrao97: Fixed.

course3/homework-list/list.h
karunrao97: Fixed. This is a remnant of me initially returning Node objects instead of iterators.

course3/homework-list/test.cc
karunrao97, line 48: Done.

course3/homework-list/list.h
karunrao97: Fixed. I somehow found it easier to implement explicit memory management in this case.

course3/homework-list/test.cc
karunrao97: Done.

karunrao97:
Thanks for the comments. Just pushed some changes - I think all the issues are fixed now, but I might've introduced new bugs.

course3/homework-list/list.h
```diff
@@ -0,0 +1,105 @@
...
+		T mData;
+		std::unique_ptr<Node> mNext;
+	};
+	
+	class iterator
+	{
+	friend class List;
+		
+	public:
+		iterator(Node* node) : mNode(node) {}
```
karunrao97, line 32: Is it okay to pass raw pointers around for iterators, or would smart pointers be better? If the latter, would the correct solution be to replace unique_ptrs with shared_ptrs, and use weak_ptrs for iterators? I thought about doing this, but didn't implement it since it seemed like overkill.

maciekgajewski, line 32: Using raw pointers is absolutely fine in this case. The use of raw pointer makes the following clear:
- The pointer is not-owning (otherwise it would be a smart pointer)
- The pointer can be null (otherwise it would be a reference)

IMO this is one of few situations when raw pointer is the appropriate type.

course3/homework-list/list.h
```diff
@@ -0,0 +1,105 @@
...
+	public:
+		Node(const T& data, std::unique_ptr<Node>& next)
+		: 	mData(data)
+		{
+			mNext = std::move(next);
+		}
+		
+	private:
+		T mData;
+		std::unique_ptr<Node> mNext;
```
maciekgajewski, line 24: I don't know how about you, but I find this much cleaner now: Node _owns_ mNext (hence the smart pointer), while iterator only passively points at it (hence the raw pointer).

course3/homework-list/list.h
```diff
@@ -0,0 +1,105 @@
...
+		bool operator!=(const const_iterator& it) { return !operator==(it); }
+		const_iterator& operator++() { mNode = mNode->mNext.get(); return *this; }
+		const T& operator*() { return mNode->mData; }
+		
+	private:
+		Node* mNode = nullptr;
+	};
+	
+	List() = default;
+	List(const std::initializer_list<T>& list) { for (const T& t : list) push_front(t); }
```
maciekgajewski, line 59: Nice

course3/homework-list/list.h
```diff
@@ -0,0 +1,105 @@
...
+	iterator end() { return nullptr; }
+	const_iterator end() const { return nullptr; }
+	
+	bool empty() const { return mSize == 0; }
+	std::size_t size() const { return mSize; }
+	
+	T& front() { return mBegin->mData; }
+	const T& front() const { return mBegin->mData; }
+	
+	void push_front(const T& data)
```
maciekgajewski, line 72: I propose the following test: create a list of unique_ptrs.

course3/homework-list/list.h
```diff
@@ -0,0 +1,105 @@
...
+	void erase(iterator it)
+	{
+		iterator e = begin();
+		if (e == it)
+		{
+			pop_front();
+			return;
+		}
+		
+		while (e.mNode->mNext.get() != it.mNode) { ++e; }
```
maciekgajewski: The brackets around `++e;` are not necessary. Yet, for readability, I would put it on a separate line

maciekgajewski:
I like it!

course3/homework-list/list.h
karunrao97, line 24: I agree. When first writing the code, it wasn't immediately clear to me which pointers should be smart, and I took the lazy way out. After thinking a bit about it, I much prefer this too now.

course3/homework-list/list.h
karunrao97, line 32: That's clear now. Thanks!

course3/homework-list/list.h
karunrao97: Will do

course3/homework-list/list.h
karunrao97, line 72: Will do. I guess it'll break since Node will try and copy data (which is a unique_ptr) in its constructor? Perhaps I should store references instead.

karunrao97:
Added a test for a list of unique_ptrs. List::push_front now accepts data by value, and data is afterwards moved around to ensure no copying takes place. In order for this to work, the caller needs to give up ownership to the unique_ptr when adding it to a list. Is this acceptable?

I found the following link quite useful when implementing this:
http://stackoverflow.com/a/8114913

course3/homework-list/list.h
karunrao97, line 59: The initializer list works well in general, but unfortunately seems to be incompatible with move semantics. As a result, I cannot use this implementation to create a list of unique_ptrs.

More info on this here (I didn't try any of the fancy workarounds):
http://stackoverflow.com/a/8468817

karunrao97:
Did some further reading up on STL containers, and it seems they always copy incoming data, which is consistent with my current implementation. However, instead of accepting data by value (as in my implementation), the STL containers accept them either by const lvalue reference or by rvalue reference (e.g. http://en.cppreference.com/w/cpp/container/vector/push_back) - what's the reason to prefer this over simply copying by value in push_back and then using std::move internally?

maciekgajewski:
This is not correct, containers not always copy. When accepting rvalue reference, they _move_, this is to allow for working with non-copyable types, like unique_ptr.

On top of that, there is the enttre family of emplace methods, that create the contained object in-place, without neither moving nor copying.

karunrao97:
True, I think I phrased my comment and question quite badly - my implementation actually also doesn't copy when an rvalue reference is provided (since it always moves the object around internally) - it also seems to work for unique_ptrs. On thinking about it a bit more, though, I see a big downside to my implementation - if an object does not implement move semantics, then the internal std::move casts aren't really moving anything, and we end up copying the object several times - for non moveable objects, a const lvalue reference would be much better. I'll implement two overloaded methods instead, similar to the STL containers.

karunrao97:
Fixed now, I think.

PR #39, opened 2016-06-12
Custom unique pointer and make_unique (with memory leak!). Added begin() and end().

wiggity:
There's a memory leak here that I can't track down :(


course3/homework-string/test.cc
```diff
@@ -93,14 +94,19 @@ void string_test()
 	assert(StringType("0") < StringType("1"));
 	assert(!(StringType("1") < StringType("0")));
 	assert(!(StringType("1") < StringType()));
+	
+	StringType forIterating("FoooBar");
+	int i(0);
+	for (auto it(forIterating.begin()); it != forIterating.end(); it++, i++) {
```
maciekgajewski, line 106: Why don't you try the range-based for?

```
for(chat c : forIterating)
```

course3/homework-string/my_unique_ptr.h
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+};
+
+template<class T>
+/* When I use std::size_t instead of int my unbounded array make_unique_ptr function
+ * isn't called, the single object function is called instead. This results in
+ * compilation errors naturally...dunno why the correct template function isn't being
+ * inferred.*/
+MyUniquePtr<T> MyMakeUnique(int num_entries)
```
maciekgajewski: I don't know the answer now, but why don't you see how std::make_unique does that?

course3/homework-string/test.cc
wiggity, line 106: done

course3/homework-string/my_unique_ptr.h
wiggity: Fixed it..dunno what I did wrong before.

PR #41, opened 2016-06-15
Linked list

dgkimpton:


dgkimpton:
I think this implementation has major issues, I'll have some questions about it in the class.

I tried to store a T& in the Nodes which of course doesn't make sense... because the referenced item can go out of scope before the list does.

What I need to do I think is store a Copy of the data in the Node, and probably implement a move constructor so that whatever data is passed to the node can be moved.

The strange thing is this requires data that is stored in this list to have a copy constructor... is that expected?

dgkimpton:
Also, I'm pretty sure unique_ptr should have played a role in here... something to try out later I guess.

course3/homework-list/linked-list.h
```diff
@@ -0,0 +1,164 @@
...
+    }
+    
+    ~LinkedList()
+    {        
+        auto current = mHead;
+        while(current != nullptr)
+        {
+            auto tmp = current;
+            current = current->Next;
+            delete tmp;
```
maciekgajewski: "Explicit `delete` should be considered a bug"

maciekgajewski:
What on earth is the 10k 'catch.h' file?
Is this a 3rd party stuff? This is not how you use 3rd party libs in your project in C++, this is not Java to bundle all required libs with your code :)

course3/homework-list/linked-list.h
dgkimpton: indeed, using unique_ptr instead occurred to me overnight.

dgkimpton:
yes, the 10k file is 3rd party stuff... I wasn't sure what the best way of referring to it was... since the license was permissive I included it so that at least everything would build.

maciekgajewski:
If you need unit testing framework, I suggest either Boost's unit testing, which should be available on every machine with boost installed, or Google's GTest

course3/homework-list/linked-list.h
```diff
@@ -0,0 +1,180 @@
...
+    }
+    
+    void push_front(const TNode& data)
+    {
+        mHead = new Node(data, mHead);
+    }
+
+	void push_front(TNode&& data)
+	{
+		// TODO: Writing this function feels wrong, is it?
```
maciekgajewski: Sure. The indentation is wrong, for one :)
Secondly, it uses explicit 'new', which should be avoided.
And last, but not least: it won't work as expected. Try creating a list of non-copyable objects, for instance a list of std::unique_ptr. It won't work with this implementation, but it should.

course3/homework-list/list-main.cc
```diff
@@ -0,0 +1,29 @@
+
+#include "linked-list.h"
+
+#include <iostream>
+#include <string>
+
+int main(int argc, char** argv)
+{
+    LinkedList<std::string> strings;
```
maciekgajewski: Would be nice to be able to say:

```
LinkedList<std::string> strings = { "A", "list", "of", "strings"};
```

course3/homework-list/build/.gitignore
```diff
@@ -0,0 +1,3 @@
+*
+!.gitignore
```
maciekgajewski, line 2: This way you will aways have the 'build' directory in your source, only to hold this .gitignore.

It would be much nicer _not_ to use in-source build, but rather create the build directory outside of the tree entirely.

course3/homework-list/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+
+cmake_minimum_required (VERSION 3.5 )
+
+set(CMAKE_CXX_FLAGS "-std=c++14 -Wall")
+
+
+find_package( Boost 1.6 COMPONENTS program_options REQUIRED)
```
dgkimpton, line 7: I was using 'program_options' to try and see if I could get boost to work, answer: Linux - easy, Windows - not yet, still fighting it.

course3/homework-list/linked-list.h
```diff
@@ -0,0 +1,241 @@
...
+
+        Iterator& operator++()
+        {
+            mPointer = mPointer->next();
+            return *this;
+        }
+
+        Iterator operator++ (int)
+        {
+            // FIXME: wrong order
```
dgkimpton, line 103: I fixed this but forgot to commit it, clearly I don't have enough unit tests yet.

course3/homework-list/list-main.cc
```diff
@@ -0,0 +1,47 @@
...
+    boost::program_options::notify(args);
+    
+    if (args.count("help")) {
+        std::cout << desc << std::endl;
+        return 1;
+    }
+    
+    LinkedList<int> v{ 2, 4, 2, 0, 5, 10, 7, 3, 7, 1 };
+
+   // std::rotate(v.begin(), v.begin()++, v.end());
```
dgkimpton, line 29: I got stuck here: I seem to be missing a 'swap' method from something in order to make the Iterator a full ForwardIterator but after hours of trying various different syntaxes and implementations I gave up. I'm hoping you can explain this in the lesson.

PR #42, opened 2016-06-15
implemented basic linked list (WIP) - Alex

neuraltoys:
still have to implement the iterator
but at the moment cannot compile and I have no idea why

I get this:
CMakeFiles\linked_list_test.dir/objects.a(test.cc.obj): In function `main':
test.cc:9: undefined reference to`LinkedList<int>::LinkedList()'
test.cc:10: undefined reference to `LinkedList<int>::empty() const'
test.cc:11: undefined reference to`LinkedList<int>::begin()'
test.cc:11: undefined reference to `LinkedList<int>::end()'
test.cc:13: undefined reference to`LinkedList<int>::push_front(int const&)'
test.cc:14: undefined reference to `LinkedList<int>::empty() const'
test.cc:15: undefined reference to`LinkedList<int>::size() const'


maciekgajewski:
The code _compiles_ fine, this is a linker error.
in main.cc you are calling `LinekdList<int>::empty()`. Compiler sees the instantiation of the class template `LinkedList<int>`, and sees the _declaration_ of the method `empty()`.

Then the linker tries to mach this call to `LinkedList<int>::empty()` to the _definition_ of the function, and fails: the function is never defined!

course3/homework-linked-list/linked_list.h
```diff
@@ -8,29 +8,66 @@ class LinkedList
...
+        this->length++;
+    }
+
+    void pop_front() {
+        if (this->empty()) {
+            throw std::out_of_range("The list is empty");
+        }
+        auto topop = this->head;
+        this->head = this->head->next;
+        delete topop;
```
maciekgajewski, line 50: Don't use explicit 'delete' or explicit 'new', they are breeding ground for bugs!

PR #43, opened 2016-06-19
Linkedlist

zerokill:
All string code should not have been part of this :/


maciekgajewski:
So what is it I should have a look in this pull request: vector, string or list?
Please, put the list in a separate branch, and create a separate pull request.

PR #44, opened 2016-06-19
Linkedlist

zerokill:
Removed string homework. Working on iterator_traits


course3/homework-linkedlist/linkedlist.h
```diff
@@ -0,0 +1,122 @@
...
+    public:
+        const_iterator(const iterator& it) : mNode(it.mNode) {}
+        const_iterator(Node* node) : mNode(node) {}
+        bool operator==(const const_iterator& it) { return mNode == it.mNode; }
+        bool operator!=(const const_iterator& it) { return mNode != it.mNode; }
+        const_iterator& operator++() { 
+            mNode = mNode->mNext.get();
+            return *this;
+        }
+        T& operator*() {
```
maciekgajewski: `const_iterator` shouldn't provide a non-const access to the content of the list.

course3/homework-linkedlist/linkedlist.h
```diff
@@ -0,0 +1,129 @@
...
+
+// Reverse iterator. http://stackoverflow.com/questions/8542591/c11-reverse-range-based-for-loop
+template <typename T>
+struct reversion_wrapper { T& iterable; };
+
+template <typename T>
+auto begin (reversion_wrapper<T> w) { return rbegin(w.iterable); }
+
+template <typename T>
+auto end (reversion_wrapper<T> w) { return rend(w.iterable); }
```
maciekgajewski, line 15: This will only work for Ts which are in the `std` namespace, as `rbegin` and `rend` are in the namespace ,and you don't specify it here. So the fact that it may work depends solely on "Argument Dependent Lookup": http://en.cppreference.com/w/cpp/language/adl

course3/homework-linkedlist/linkedlist.h
```diff
@@ -0,0 +1,129 @@
+#pragma once
+#include <iostream>
+#include <memory>
+
+namespace mau {
+
+// Reverse iterator. http://stackoverflow.com/questions/8542591/c11-reverse-range-based-for-loop
+template <typename T>
+struct reversion_wrapper { T& iterable; };
```
maciekgajewski, line 9: Interesting! Good idea.
But you could've use `boost::adaptors::reversed` : http://www.boost.org/doc/libs/1_58_0/libs/range/doc/html/range/reference/adaptors/reference/reversed.html

course3/homework-linkedlist/linkedlist.h
```diff
@@ -0,0 +1,129 @@
...
+            return mNode->mData;
+        }
+
+        Node* mNode = nullptr;
+    };
+
+    list() : mSize() {}
+
+    list(std::initializer_list<T> init) : mSize(0) {
+        for(auto i : reverse(init))
```
maciekgajewski: or:

```
for(const auto& i : init | boost::adaptors::reversed)
    push_front(i);
```

maciekgajewski: You should rather use `const auto& i` or `const T&`, otherwise `i` will be an unnecessary extra copy of the list's element.

course3/homework-linkedlist/linkedlist.h
```diff
@@ -0,0 +1,129 @@
...
+            push_front(i);
+
+    }
+
+    iterator begin() { return mHead.get(); }
+    const iterator begin() const { return mHead.get(); }
+    iterator end() { return nullptr; }
+    const iterator end() const { return nullptr; }
+
+    std::size_t size() {
```
maciekgajewski: should be const

course3/homework-linkedlist/linkedlist.h
```diff
@@ -0,0 +1,129 @@
...
+        return mSize;
+    }
+
+    bool empty() {
```
maciekgajewski: should be const

course3/homework-linkedlist/linkedlist.h
```diff
@@ -0,0 +1,129 @@
...
+        T& operator*() {
+            return mNode->mData;
+        }
+
+        Node* mNode = nullptr;
+    };
+
+    list() : mSize() {}
+
+    list(std::initializer_list<T> init) : mSize(0) {
```
maciekgajewski: This is fine, but this will only be selected for initializer_lists of objects of exactly the type `T`.
Think of the following use-cases:

```
mau::list<std::string> ls = {"abc", "def"}; // T=std::string, but initialized with std::initializer_list<const char*>
mau::list<double> ld = {1, 3, 5}; // T=double, but initialized with std::initializer_list<double>
```

zerokill: I need to have a think about this. I have created a failing testcase and I see the problem!

PR #45, opened 2016-06-26
list algorithms + benchmark

karunrao97:


course3/homework-list/list.h
```diff
@@ -0,0 +1,116 @@
...
+
+
+namespace Karun
+{
+
+template<typename T>
+class List
+{
+public:
+	typedef T value_type;
```
karunrao97: This was required in order for std::front_inserter to work. Is there a better way?

course3/homework-list/test.cc
```diff
@@ -0,0 +1,170 @@
...
+	}
+	
+	// distance
+	l = {1, 1, 1, 1, 2, 2, 2, 3, 3, 4};
+	assert(std::distance(l.begin(), l.end()) == 10);
+	
+	// count and count_if
+	assert(std::count(l.begin(), l.end(), 2) == 3);
+	assert(std::count_if(l.begin(), l.end(), [](int n){ return n >= 2; }) == 6);
+	//assert(boost::count(l, 2) == 3);
```
karunrao97: The code doesn't compile for Boost range based algorithms. Any hints on what's required?

course3/homework-list/benchmark.cc
```diff
@@ -0,0 +1,24 @@
...
+
+template<template<typename...> class ListType>
+void ListBenchmark(benchmark::State& state)
+{
+	ListType<int> l;
+	for (int i=0; i < state.range_x(); ++i)
+		l.push_front(i);
+}
+
+//BENCHMARK_TEMPLATE(ListBenchmark, Karun::List)->Range(1<<0, 1<<10);
```
karunrao97: BENCHMARK_TEMPLATE and BENCHMARK_MAIN don't compile. I get an "Undefined symbols for architecture x86_64" error. I'm running the code on OS X with clang++.

course3/homework-list/list.h
maciekgajewski: No. This is the way. Actually, I personally prefer the "using" version:

```
using value_type = T;
```

Some code needs these to infer stored type (and other types) from the container type.

course3/homework-list/test.cc
maciekgajewski: The main requirements are 'begin' and 'end' and well-behaving iterators. I can't say what's wrong without seeing the error message.

course3/homework-list/benchmark.cc
maciekgajewski: I can't see any inking with the benchmark library in your CMakeLists.txt

course3/homework-list/test.cc
karunrao97: Error message attached.
[error.txt](https://github.com/maciekgajewski/c-excercises/files/335649/error.txt)

course3/homework-list/benchmark.cc
karunrao97: I added the target_link_libraries (see updated CMakeLists.txt), but now get the following linker error.
[error.txt](https://github.com/maciekgajewski/c-excercises/files/335664/error.txt)

maciekgajewski: The error is pretty clear: `library not found for -lbenchmark`
Do you have the library installed in a place where the linker can find it? On Linux, I'd be looking for a file named `libbenchmark.a` or `libbenchmark.so.X.Y` in one of the standard system paths. I don't know what the convention is on your OS.

course3/homework-list/test.cc
maciekgajewski: I must commend the clang/boost combo for producing really readable error messages:

```
error: cannot increment value of type 'Karun::List<int>::iterator'
                (void)i++;
                      ~^
```

Your iterator class doesn't provide _postfix_ increment operator. I don't know why boost wants to use postfix (`i++`) instead if prefix (`++i`) version. but if they require it, then let's provide it.

http://en.cppreference.com/w/cpp/language/operators

karunrao97: Thanks, fixed. Boost algos working now :)

course3/homework-list/benchmark.cc
karunrao97: I have `libbenchmark.a` in `/usr/local/lib`, which is in my `$PATH`. Is there an `lbenchmark` file also expected somewhere?

karunrao97: Fixed by adding `/usr/local/lib` to `$LIBRARY_PATH` instead.

karunrao97:
Benchmarks "work" now (see output here: [bench.txt](https://github.com/maciekgajewski/c-excercises/files/337796/bench.txt))

Unfortunately, as you can see in the output, my list implementation segfaults for large lists. Turns out it segfaults as soon as we get to exactly 100k elements. Any idea why this magic number?

maciekgajewski:
Is your benchmark build in debug or release mode?

maciekgajewski:
Do you have a core file from the crash? Have you tried running it in debugger to see hwere exactly the crash happens?

maciekgajewski:
I suspect that the recursive destruction of elements runs out of stack. There are 2 solutions: compile it in release mode and hope that compiler will replace the recursion with iteration, or manually write recursion-free destructor.

karunrao97:
I tried compiling in release mode, but it still failed. I then implemented an iterative destructor, and it works fine! Thanks :)

karunrao97:
Updated benchmark results (for `push_front`, `copy`, `move` and `transform`) here: [bench.txt](https://github.com/maciekgajewski/c-excercises/files/338098/bench.txt)

In general, performance is similar to `std::forward_list`, slightly better than `std::list`, and far worse than `std::deque`.

maciekgajewski:
Great job!

PR #47, opened 2016-06-29
transcoder

karunrao97:
A rough first attempt - no unit tests yet, but a simple functional test instead. 

The transcoder can read/write both JSON and XML files, so conversion in either direction is possible. Interestingly, there seems to be an issue converting JSON lists to XML, since ptree is unable to parse output.xml, which is generated by converting input.json to output.xml. When the list elements in output.xml are removed (this is what I did in order to create input.xml), ptree is able to parse the XML file.

### DIFF
--- a/course3/homework-transcoder/CMakeLists.txt
+++ b/course3/homework-transcoder/CMakeLists.txt
@@ -5,5 +5,5 @@ find_package(Boost REQUIRED)
 include_directories(${Boost_INCLUDE_DIRS})
 
 add_executable(test
-	transcoder.h test.cc
+	transcoder.h transcoder.cc test.cc
 )
--- a/course3/homework-transcoder/CMakeLists.txt
+++ b/course3/homework-transcoder/CMakeLists.txt
@@ -1,0 +1,9 @@
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -std=c++14")
+
+find_package(Boost REQUIRED)
+
+include_directories(${Boost_INCLUDE_DIRS})
+
+add_executable(test
+	transcoder.h test.cc
+)
--- a/course3/homework-transcoder/test.cc
+++ b/course3/homework-transcoder/test.cc
@@ -1,0 +1,16 @@
+#include "transcoder.h"
+
+#include <iostream>
+
+
+void transcoder_test()
+{
+	
+}
+
+int main()
+{
+	std::cout << "Testing Karun::Transcoder..." << std::endl;
+	transcoder_test();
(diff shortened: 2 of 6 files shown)


course3/homework-transcoder/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -std=c++14")
+
+find_package(Boost REQUIRED)
```
maciekgajewski, line 3: Use boost filesystem library. This is one of the few boost libs you have to link with

course3/homework-transcoder/transcoder.cc
```diff
@@ -0,0 +1,38 @@
...
+#include <string>
+#include <exception>
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/xml_parser.hpp>
+#include <boost/property_tree/json_parser.hpp>
+
+
+Karun::Transcoder::Transcoder(const std::string& fileName)
+{
+	if (fileName.substr(fileName.size() - 4) == ".xml")
```
maciekgajewski, line 12: Boost.Filesystem has a function to get file extension

course3/homework-transcoder/transcoder.h
```diff
@@ -0,0 +1,25 @@
...
+	Transcoder() = default;
+	Transcoder(const std::string& fileName);
+	
+	void ReadXML(const std::string& fileName);
+	void ReadJSON(const std::string& fileName);
+	void WriteXML(const std::string& fileName);
+	void WriteJSON(const std::string& fileName);
+	
+private:
+	boost::property_tree::ptree mData;
```
maciekgajewski, line 22: Does it really have to be a class? There is no much state in it, and this could be much simple implemented as a pair of functions

course3/homework-transcoder/test.cc
```diff
@@ -0,0 +1,17 @@
+#include "transcoder.h"
+
+#include <iostream>
+
+
+int main()
+{
+	std::cout << "Transcoding input.json to output.xml..." << std::endl;
```
maciekgajewski, line 8: This is fine for a test driver for your library, but lets try an make it into a fully-usable, Unix-style tool.
This means that the following invocations should be possible:

```
$ transcoder --xml2json in.json out.xml
$ transcoder --help # prionts help and exists
$ transcoder -v --json2xml in.xml /dev/stdout # p-v prints extra debugging information to STDERR
```

Use Boost.ProgramOptions to parse command-line params

PR #49, opened 2016-07-03
Added hash function/support for unordered map and set.

wiggity:


course3/homework-string/my_string.h
```diff
@@ -1,6 +1,49 @@
...
+		size_t size() const;
+		const char* c_str() const;
+		const char* begin() const;
+		const char* end() const;
 };
+
+/*
+ * Don't understand why I need to wrap this with namespace std :(
+ * If I don't I get the following compilation error:
+ * /home/wshayes/projects/cpp_course/c-excercises/course3/homework-string/my_string.h:46:9: error: ‘hash’ is not a class template
```
maciekgajewski, line 41: You can create your own functor, in any namespace you like, and pass it as a template parameter to the container.
Is that by default the containers will use `std::hash<Key>`

PR #50, opened 2016-09-17
homework 1: operator overload

CamiloOMesa:


course4/homework1/vector2d.h
```diff
@@ -2,8 +2,29 @@
 
 #include <ostream>
 
+
 struct Vector2d
 {
-	// TODO
+  int X;
```
maciekgajewski: Bad naming. ALL CAPS names are usually used for constants, in most of the coding styles out there.

course4/homework1/vector2d.h
```diff
@@ -2,8 +2,29 @@
...
+  {
+    Vector2d r;
+    r.X = X + a.X;
+    r.Y = Y + a.Y;
+    return r;
+  }
+  
+  bool operator==(const Vector2d& a)
+  {
+    return X == a.X and Y == a.Y;
```
maciekgajewski: Don't use the textual form of 'and', use '&&' instead. Some compilers get stupid when they see 'and', and some libraries define macro/function with that name.

course4/homework1/vector2d.h
```diff
@@ -2,8 +2,29 @@
...
+  }
 };
 
+  std::ostream& operator<<(std::ostream& ss, const Vector2d& a)
```
maciekgajewski, line 63: Both the above methods should be 'const'

course4/homework1/test.cc
```diff
@@ -6,17 +6,17 @@
...
 #ifdef NDEBUG
 #error Compile the code in debug mode!
 #endif
 
-	std::cout << "Starting test..." << std::endl;	
+        std::cout << "Starting test..." << std::endl;
 
-	Vector2d a = {5, 6};
-	Vector2d b = {-1, 7};
+        Vector2d a = Vector2d::Rectangular(5, 6);
```
maciekgajewski, line 16: You changed the indentation from tabs to spaces. 
It is OK to use any indentation scheme that suits you, but changing ti in existing code only creates "whitespace-only" changes in a patch, which increase noise-to-signal ratio.
Also, if this would be "real" pull request, I would probably reject it fro messing with my indents

course4/homework1/vector2d.h
```diff
@@ -1,9 +1,69 @@
...
+  int mY;
+  Vector2d(int x, int y):mX(x),mY(y) {}
+
+public:
+  static Vector2d Rectangular(int x, int y)
+  {
+      return Vector2d(x,y);
+  }
+
+  static Vector2d Polar(double r, double degrees)
```
maciekgajewski, line 18: Why using degrees in your API, when all the other math functions use radians?

course4/homework1/vector2d.h
```diff
@@ -1,9 +1,69 @@
...
+  {
+      double radians = degrees*M_PI/180;
+      int x = r * cos(radians);
```
maciekgajewski, line 21: Please, use qualified version of these functions: `std::cos`, `std::sin` etc

PR #51, opened 2016-09-18
homework one 

arccth:
By Oleksii Hidlevskyi

### DIFF
--- a/course4/homework1/vector2d.h
+++ b/course4/homework1/vector2d.h
@@ -4,6 +4,41 @@
 
 struct Vector2d
 {
-	// TODO
+private:
+   double mX, mY;
+
+public:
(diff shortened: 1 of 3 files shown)


course4/homework1/vector2d.h
```diff
@@ -4,6 +4,41 @@
 
 struct Vector2d
 {
-	// TODO
+private:
+   double mX, mY;
+
+public:
+
+   Vector2d(const double& x, const double& y)
```
maciekgajewski: There is no value in passing double by const ref. Copying double and copying a pointer is exactly the same operation: 8-byte copy.

maciekgajewski (commented):
Good!

PR #54, opened 2016-09-22
Ningyao finished homework1

AoyamaG:
Please take a look and leave some comments.

### DIFF
--- a/course4/homework1/CMakeLists.txt
+++ b/course4/homework1/CMakeLists.txt
@@ -1,0 +1,3 @@
+project(homework1)
+include_directories(.)
+add_executable(test test.cc vector2d.h)
--- a/course4/homework1/CMakeLists.txt
+++ b/course4/homework1/CMakeLists.txt
@@ -1,3 +1,0 @@
-project(homework1)
-include_directories(.)
-add_executable(test test.cc vector2d.h)
--- a/course4/homework1/vector2d.h
+++ b/course4/homework1/vector2d.h
@@ -5,5 +5,23 @@
 struct Vector2d
 {
 	// TODO
+private:
+	double x;
+	double y;
+public:
+	Vector2d() = delete;
+	Vector2d(double a, double b) : x(a), y(b) {}
+	Vector2d operator+ (const Vector2d& v2) 
+	{
+		return{ x + v2.x, y + v2.y };
+	};
+	bool operator== (const Vector2d& v2)
+	{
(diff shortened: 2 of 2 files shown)


course4/homework1/vector2d.h
```diff
@@ -5,5 +5,23 @@
 struct Vector2d
 {
 	// TODO
+private:
+	double x;
+	double y;
+public:
+	Vector2d() = delete;
```
maciekgajewski: Why have you decided to do it?

course4/homework1/vector2d.h
```diff
@@ -5,5 +5,23 @@
...
+	Vector2d(double a, double b) : x(a), y(b) {}
+	Vector2d operator+ (const Vector2d& v2) 
+	{
+		return{ x + v2.x, y + v2.y };
+	};
+	bool operator== (const Vector2d& v2)
```
maciekgajewski: Both t== and + operators should be const

course4/homework1/vector2d.h
```diff
@@ -5,5 +5,23 @@
...
+	{
+		return (x == v2.x) && (y == v2.y) ? true : false;
```
maciekgajewski: The ? : operation doesn't add anything; it makes no sense to use it.

course4/homework1/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+project(homework1)
+include_directories(.)
```
maciekgajewski: This should never be needed

maciekgajewski (commented):
Mostly correct. There is few weird, redundant and unnecessary bits, which should be removed.

course4/homework1/vector2d.h
AoyamaG: Agree. Do not know why I used ?:

course4/homework1/vector2d.h
AoyamaG: I did not want a vector being initialized with random numbers.

course4/homework1/CMakeLists.txt
AoyamaG: I use visual studio. 
project(homework1) creates a solution file for me to open up using vs.
The second line I agree I should delete.

maciekgajewski: "project" is fine, the "include_directories" is not needed.

PR #55, opened 2016-09-27
Implementation of Vector2d and polar coordinates

florian-kloeck:


course4/homework1/test.cc
```diff
@@ -22,6 +22,9 @@ int main()
 	s << a;
 
 	assert(s.str() == "(5, 6)");
+	
+	double pi = 3.14159265358979323846;
```
maciekgajewski: You may try using M_PI from <cmath>

course4/homework1/test.cc
```diff
@@ -22,6 +22,9 @@ int main()
...
+	std::cout << Vector2d(PolarCoords2d(5, pi/2)) << std::endl;
```
maciekgajewski: This is not really a test. It just prints the value., and requires manual verification of its correctness.
Use `assert` to make the test fully automated.

course4/homework1/vector2d.h
```diff
@@ -1,9 +1,68 @@
...
+#include <ostream>
+#include <cmath>
+
+struct PolarCoords2d;
+
+struct CartesianCoords2d
+{
+	double x, y;
+	
+	CartesianCoords2d(double x, double y) : x(x), y(y) {}
```
maciekgajewski: Since the struct is POD, you don't really need the constructor, you can just create instance using list initialization:
`CartesianCoords2d{1.2, 3.4}`

course4/homework1/vector2d.h
```diff
@@ -1,9 +1,68 @@
...
+
+	bool operator==(const Vector2d& other_vec) const{
+		return coords==(other_vec.coords);
+	}
+	
+	CartesianCoords2d get_cartesian_coords() const{
+		return coords;
+	}
+
+	PolarCoords2d get_polar_coords() const{
```
maciekgajewski, line 60: Interesting design, but it is not the most efficient one. 
If I'm interested in `r` only, I still need to get the full set of polar coordinates, and that means calculating `phi` as well.
This violates the principle of "not paying for what you are not using"

course4/homework1/vector2d.h
```diff
@@ -1,9 +1,68 @@
...
+		return coords.get_polar_coords();
+	}
+
+	private:
+		CartesianCoords2d coords;
+};
+
+std::ostream& operator<<(std::ostream& s, const Vector2d& vector){
+		return s << '(' << vector.get_cartesian_coords().x << ", " << vector.get_cartesian_coords().y << ')';
```
maciekgajewski: and here,you get the full copy of cartesian coords twice. Unnecessary overhead.

maciekgajewski (commented):
There is no tests for the new functionality

course4/homework1/test.cc
florian-kloeck: Fully automated testing requires handling the rounding errors correctly, but I'll do that

course4/homework1/vector2d.h
florian-kloeck: Yes but only if you have a use case where you need r or phi separately. I chose this design over the named constructors because the 2 doubles have a different meaning if they represent polar coordinates or Cartesian coordinates, and this makes the different meaning explicit by using different types for them.

course4/homework1/vector2d.h
florian-kloeck: True, I checked the assembly code gcc generates and the function is indeed called twice. Why can't or doesn't the compiler optimize this away? In the end vector is const reference and get_cartesian_coords is a const member function, so the result needs to be the same for all function calls? Or doesn't the compiler know that the other function calls here cannot have side effects on vector?

course4/homework1/test.cc
florian-kloeck: Compiler complains that M_PI is not defined, both if i include cmath or math.h - also using the obscure #define _USE_MATH_DEFINES before. This is with gcc 5.4.0

course4/homework1/vector2d.h
maciekgajewski: You are not really using different types for the doubles, but rather for structures containing them. Indeed, having a separate type for "length" and "angle" would be the cleanest solution, the one you are proposing is something else.

Also: why are you assuming that rho and phi are always needed together? It's very easy to imagine a program that only ever uses one of them.

course4/homework1/vector2d.h
maciekgajewski: Probably, if you would compile it in release mode, compiler could have optimized it. But this is not the kind of optimization that I would relay on; the calls to std::sin, std::cos etc may have side effects  the compiler doesn't know of, so it may decided not to elide them away.

PR #56, opened 2016-09-29
Homework 1 - Vector2d

ch-iel:
Slightly delayed pull request. The compilation issues I had before were due to Make using an old compiler (<c++11).


maciekgajewski (commented):
Good

PR #57, opened 2016-10-04
string homework

CamiloOMesa:


maciekgajewski (commented):
Please, create separate pull request per homework assignment

PR #58, opened 2016-10-05
string homework

CamiloOMesa:


course4/homework-string/string.hh
```diff
@@ -1,10 +1,35 @@
...
 {
-	string();
+public:
+    string();
+    ~string();
+    string(const char* chars);
+    string(const string& source);
+
+    string& operator=(const string& source);
+    bool operator ==(const string& a);
```
maciekgajewski, line 15: the comparison operators should be const

course4/homework-string/string.hh
```diff
@@ -1,10 +1,35 @@
...
+    bool operator !=(const string& a);
+    char& operator[](int index) const;
+
+    bool empty() const;
+    int size() const;
+
+private:
+    std::unique_ptr<char[]> mCharArray;
+    int mLength;
+    int getLenghtOfChars(const char* chars);
```
maciekgajewski, line 26: Don't reinvent the wheel, use existing functions from <cstring>

course4/homework-string/string.hh
```diff
@@ -1,10 +1,35 @@
...
+    void deepCopy(const char* chars, int lenght);
+    bool deepCompare(const char* chars, int length);
 };
 
-}
\ No newline at end of file
+std::stringstream& operator <<(std::stringstream& os, const course::string &s);
```
maciekgajewski, line 31: wrong! This will only worj with stringstresam, but not with std::cout or std::fstream. Use `std::ostream`, it's the base class.

course4/homework-string/string.cc
```diff
@@ -1,9 +1,117 @@
...
+}
+
+string::string(const string& source)
+{
+    mLength = source.size();
+    mCharArray = std::make_unique<char[]>(mLength);
+    deepCopy(source.mCharArray.get(), mLength);
+}
+
+char& string::operator[](int index) const
```
maciekgajewski, line 26: With this implementation, you are allowing for changing the content of const object. You are breaking the promise of constness.

course4/homework-string/string.cc
```diff
@@ -1,9 +1,117 @@
...
+
+    return false;
+}
+
+int string::size() const
+{
+    return mLength;
+}
+
+ int string::getLenghtOfChars(const char* chars)
```
maciekgajewski, line 76: even if we imagine that there is no `std::strlen`, then this method should `static`, it doesn't need `this`.

course4/homework-string/string.cc
```diff
@@ -1,9 +1,117 @@
...
+     for(int j=0;j<length;++j)
+     {
+         if(mCharArray[j] != chars[j])
+             return false;
+     }
+
+     return true;
+ }
+
+string::~string()
```
maciekgajewski, line 105: If not needed, then just remove it.

maciekgajewski (commented):
Mostly good, violations of const-correctness.

You decided to store the data without the zero-terminator. This is not a bad choice, but it makes implementation of some functions (for instance operator<<) more difficult.

PR #59, opened 2016-10-05
String implementation

arccth:


course4/homework1/vector2d.h
```diff
@@ -1,9 +1,71 @@
...
 
 struct Vector2d
 {
-	// TODO
+private:
+   double mX, mY;
+
+public:
+
+   Vector2d(const double x, const double y):
```
maciekgajewski, line 13: Please, deliver the string homework in new pull request.

maciekgajewski (commented):
Please, use separate pull request for new homework.

course4/homework1/vector2d.h
arccth: But it is separate, just branched from homework 1. First pull request here: https://github.com/maciekgajewski/c-excercises/pull/51

maciekgajewski: It is not separate. It is not a string homework submission if there are vector changes in the patch.

PR #60, opened 2016-10-09
vector based string implementation

arccth:
clean pull request this time

### DIFF
--- a/course4/homework-string/string.cc
+++ b/course4/homework-string/string.cc
@@ -6,4 +6,76 @@ string::string()
 {
 }
 
+string::string(const char * c)
+{
+    int i = 0;
+    while(c[i] != '\0') // I guess we are not supposed to use strlen
+    {
+        chars.push_back(c[i]);
+        ++i;
+    }
+}
+
+string::string(const string& source)
+{
+    chars = source.chars;
+}
+
+string::string(string&& source)
+{
+    chars = std::move(source.chars);
+}
+
+bool string::empty() const
+{
+    return chars.empty();
+}
+
+int string::size() const
+{
+    return chars.size();
+}
+
+bool string::operator == (const string& other) const
+{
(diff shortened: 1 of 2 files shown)


course4/homework-string/string.cc
```diff
@@ -6,4 +6,76 @@ string::string()
 {
 }
 
-}
\ No newline at end of file
+string::string(const char * c)
+{
+    int i = 0;
+    while(c[i] != '\0') // I guess we are not supposed to use strlen
```
maciekgajewski, line 12: You are _expected_ to use `std::strlen`

course4/homework-string/string.cc
```diff
@@ -6,4 +6,76 @@ string::string()
...
+}
+
+string::string(const string& source)
+{
+    chars = source.chars;
+}
+
+string::string(string&& source)
+{
+    chars = std::move(source.chars);
```
maciekgajewski, line 26: Is there a point implementing these constructors? The default implementations would do exactly the same.

course4/homework-string/string.hh
```diff
@@ -1,10 +1,47 @@
...
+
+    char& operator [](const int index);
+
+    string& operator =(const string& source);
+
+    string& operator=(string&& source);
+
+    friend std::ostream& operator<<(std::ostream& os, const string& s)
+    {
+        for(int i=0; i < s.chars.size(); ++i)
```
maciekgajewski, line 39: Use range-based for

maciekgajewski (commented):
It is, indeed, too easy.
And your implementation would be slower than a simple, unique_ptr-based one. Why?

arccth:
Created unique_ptr implementation here: https://github.com/maciekgajewski/c-excercises/pull/61
I guess it will be slower because multiple push to vector is more expensive than allocation of known size.

PR #61, opened 2016-10-09
unique ptr string implementation

arccth:


course4/homework-string/string.cc
```diff
@@ -1,9 +1,98 @@
...
+string& string::operator =(string&& source)
+{
+   stealChars(std::move(source));
+}
+
+void string::stealChars(string&& source)
+{
+    mSize = source.mSize;
+    chars = std::move(source.chars);
+    source.chars = nullptr;
```
maciekgajewski: shouldn't be needed. 'source.chars' is supposed to be a throw-away, so it shouldn't matter what it contains after the move.

course4/homework-string/string.cc
```diff
@@ -1,9 +1,98 @@
...
+
+char& string::operator [](const int index)
+{
+    return chars[index];
+}
+
+string& string::operator =(const string& source)
+{
+   mSize = source.mSize;
+   copyChars(source.chars.get());
```
maciekgajewski, line 100: Inconsistent design: stealChars doesn't set mSize, stealChars does

maciekgajewski (commented):
Good!
Now try merging with my master, and implement the additional methods.

course4/homework-string/string.cc
arccth: Done

course4/homework-string/string.cc
```diff
@@ -1,9 +1,145 @@
...
+
+int string::length() const
+{
+    return mSize;
+}
+
+void string::clear()
+{
+    mSize = 0;
+    chars = std::make_unique<char[]>(1);
```
maciekgajewski: It would be better to avoid this allocation. Seems wasteful to perform expensive operation just for one byte; and a one that never changes.

arccth: Done

arccth:
Added begin/end implementation

course4/homework-string/string.hh
```diff
@@ -1,10 +1,73 @@
...
+
+    string& operator += (const string& other);
+
+    char * begin();
+
+    char * end();
+
+    char * begin() const;
+
+    char * end() const;
```
maciekgajewski: the const version should return `const char*`, otherwise user would be able to modify the content of const object.

course4/homework-string/test.cc
```diff
@@ -120,16 +121,31 @@ void test2()
 	assert(string4.length() == 0);
 }
 
+template<typename StringType>
+void iteratorTest()
+{
+  //const iteration test
+  const StringType s("ABCBB");
+  int b_num = std::count_if(s.begin(), s.end(), [](const char& c){return c=='B';});
```
maciekgajewski: the purpose of passing params by const ref is to avoid copying of _expensive to copy_ object. On case of simple types, like `char` (aka 8-bit integer) it has no point, as the copying is cheap.

maciekgajewski (commented):
Good. Cost-correctness violated.

course4/homework-string/string.hh
arccth: Done

PR #62, opened 2016-10-10
Homework3

AoyamaG:
string homework


maciekgajewski (commented):
I'm sorry, but you are working against course3 homewrok, instead of course4.
Please, implement the string homework in course4/

PR #63, opened 2016-10-10
Homework string

AoyamaG:


course4/homework-string/string.cc
```diff
@@ -4,6 +4,129 @@ namespace course {
 
 string::string()
 {
+	mStr = std::make_unique<char[]>(1);
+	mStr[0] = NULL;
```
maciekgajewski, line 8: Don't use `NULL`, use `nullptr` instead.

course4/homework-string/string.cc
```diff
@@ -4,6 +4,129 @@ namespace course {
 
 string::string()
 {
+	mStr = std::make_unique<char[]>(1);
```
maciekgajewski, line 7: While correct, this is unpleasant to the user. Well-behaving classes should not do anything expensive in the default constructor.

course4/homework-string/string.cc
```diff
@@ -4,6 +4,129 @@ namespace course {
...
+	mStr[0] = NULL;
+	mStrSize = 0;
 }
 
-}
\ No newline at end of file
+string::string(size_t n)
```
maciekgajewski, line 12: What's the content of the string after the constructor has been called? Why would anyone use it? Consider making it explicit.

course4/homework-string/string.cc
```diff
@@ -4,6 +4,129 @@ namespace course {
...
+{
+	mStr = std::make_unique<char[]>(n + 1);
+	mStr[n] = NULL;
+	mStrSize = n;
+}
+
+string::string(const char* s1)
+{
+	mStrSize = strlen(s1);
```
maciekgajewski, line 21: Use `std::strlen` instead of `strlen`. This also applies to other library functions: prefer C++ functions over the C ones.

course4/homework-string/string.hh
```diff
@@ -1,10 +1,38 @@
 #pragma once
+#include <memory>
+#include <string.h>
```
maciekgajewski, line 3: You don't use string.h and stdexcept in this file, therefore you shouldn't include them here. This will only increase the compilation time for the users of the class.

maciekgajewski (commented):
Mostly good, you are using some C-stuff; try using C++ instead.
This "from integer" constructor creates dangerous and confusing API.

PR #65, opened 2016-10-13
Homework: Implementation of course::string

florian-kloeck:


maciekgajewski (commented):
this is based on course3/homework-string. Please, use course4/homewrok-string as your base

PR #67, opened 2016-10-17
Implementation of course::string

florian-kloeck:


course4/homework-string/string.hh
```diff
@@ -1,10 +1,110 @@
...
+#include <cstring>
+#include <memory>
+
 namespace course {
 
 class string
 {
-	string();
+public:
+	string() : string("") {}
```
maciekgajewski: So the default constructor will cause heap allocation? Can you try to avoid it? It's much nicer to have a cheap default ctor.

course4/homework-string/string.hh
```diff
@@ -1,10 +1,110 @@
...
+	string(const char* init_str) : my_ptr(new char[std::strlen(init_str)+1])
```
maciekgajewski: Please, use make_unique instead of raw "new"

course4/homework-string/string.hh
```diff
@@ -1,10 +1,110 @@
...
+	{
+		std::strcpy(my_ptr.get(), init_str);
+	}
+	string(const string& other_string) : my_ptr(new char[other_string.size()+1])
+	{
+		std::strcpy(my_ptr.get(), other_string.my_ptr.get());
+	}
+	string(const char* init_str1, const char* init_str2) : my_ptr(new char[std::strlen(init_str1)+std::strlen(init_str2)+1])
```
maciekgajewski: why would you provide this method as a part of the public interface?

course4/homework-string/string.hh
```diff
@@ -1,10 +1,110 @@
...
+	{
+		std::strcpy(my_ptr.get(), init_str1);
+		std::strcat(my_ptr.get(), init_str2);
+	}
+
+	bool empty() const
+	{
+		return size()==0;
```
maciekgajewski: What is the algorithmic complexity of this implementation of `empty` for non-empty string?
(note: this string implementation is never actually empty)

course4/homework-string/string.hh
```diff
@@ -1,10 +1,110 @@
...
+	}
+
+	int length() const
+	{
+		return size();
+	}
+
+	void clear()
+	{
+		char* new_ptr = new char[1];
```
maciekgajewski: Do not use raw pointers and raw new.

maciekgajewski (commented):
The implementation is correct, but has some style and performance issues:
- Don't use raw pointer and raw new/delete
- Default constructor should be as cheap as possible; currently it performs heap allocation; may be super slow.
- empty() should be O(1), is O(N)

florian-kloeck:
I tried to denote an empty string with setting the pointer to nullptr. Unfortunately I ran into two issues:
- string.hh line 97 gives a compiler warning now. I understand why but I don't see a good way around that. One possibility would be to let the pointer point to an empty string as soon as this gets invoked, but I think this is really ugly.
- test.cc line 69 gives a seg fault and I got no clue why. Unfortunately I run Windows and apparently I can't use valgrind on Windows.

florian-kloeck:
The seg fault came from missing #include <iostream> in string.hh. I find it very odd that this does not produce a compiler warning/error but seg fault.
Fixed a small nasty bug if left hand side of += is empty string.
Fixing c_str() was actually very easy, just replaced the return value from char\* to const char*.

PR #68, opened 2016-10-19
Unique pointer

CamiloOMesa:


course4/SmartPointer/smart.hh
```diff
@@ -0,0 +1,59 @@
...
+        delete mPtr;
+        mPtr = nullptr;
+    }
+
+    T& operator*()
+    {
+        return *mPtr;
+    }
+
+    T* operator->()
```
maciekgajewski, line 41: const versions of these operators would be useful

course4/SmartPointer/test.cc
```diff
@@ -0,0 +1,59 @@
...
+     UP up1b(nullptr);  // up1b is empty
+
+     std::cout << "test constructor (2)...\n";
+     {
+         UP up2(new Foo); //up2 now owns a Foo
+     } // Foo deleted
+
+     std::cout << "test move...\n";
+     {
+        UP up5a(new Foo);
```
maciekgajewski, line 31: Would be better to have something like make_unique

maciekgajewski (commented):
Good!
- make_unique needed
- it would better if the test was automated; you cna use static/global variables to count instances of Foo

PR #70, opened 2016-11-02
implementation of linked list and tests

florian-kloeck:


course4/homework-linked-list/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
+
+set(CMAKE_CXX_STANDARD 14)
```
maciekgajewski, line 3: This is nicer than the explicit setting of params that I usually do. Unfortunately, it doesn't enable warnings.

course4/homework-linked-list/linked_list.hh
```diff
@@ -0,0 +1,106 @@
...
+
+namespace course {
+
+template<typename T>
+struct pointer_value_pair
+{
+	pointer_value_pair(std::unique_ptr<pointer_value_pair<T>>&& pointer, T& value) : value(value), pointer(std::move(pointer)) {}
+	
+	std::unique_ptr<pointer_value_pair<T>> pointer;
+    T& value;
```
maciekgajewski: Something's wrong with indentation here

course4/homework-linked-list/test.cc
```diff
@@ -0,0 +1,82 @@
...
+	assert(std::find_if(linked_list.begin(), linked_list.end(), [&](int n){return n>=i1;}) == linked_list.begin());
+	assert(*std::find_if(linked_list.begin(), linked_list.end(), [&](int n){return n>=i2;}) == ((i2>i1)?i2:i1));
+	assert(std::find_if(linked_list.begin(), linked_list.end(), [&](int n){return n>std::max(i1, i2);}) == linked_list.end());
+	assert(empty.begin()==empty.end());
+	assert(std::find_if(empty.begin(), empty.end(), [&](int n){return n>std::max(i1, i2);}) == empty.end());
+
+	//changing the value in the front of the list
+	linked_list.front() = i2;
+	assert(linked_list.front() == i2);
+}
```
maciekgajewski, line 87: Please, try the following test: fill the list with 100 numbers, in a loop, calculate the sum using std::accumulate

maciekgajewski (commented):
Your list, while mosty good, is not really a container.
Please, expand the tests as suggested.

maciekgajewski (commented):
Good!

course4/homework-linked-list/count_until.hh
```diff
@@ -0,0 +1,20 @@
...
+#include <iterator>
+
+namespace course {
+
+template<class InputIt, class T>
+typename std::iterator_traits<InputIt>::difference_type
+    count_until(InputIt first, InputIt last, const T& value, const T& stop)
+{
+    typename std::iterator_traits<InputIt>::difference_type ret = 0;
+    for (; (first != last) && (*first)!=stop; ++first) {
```
maciekgajewski, line 12: Can you make it _without_ using the for-loop? (or any other loop)

maciekgajewski (commented):
Try implementing the algo w/o explicit loop

course4/homework-linked-list/count_until.hh
florian-kloeck: I can think of two solutions:
- std::count(first, std::find(first, last, stop), value);
- or using std::count_if with a predicate that returns false after observing the stop value

but both of these have worse complexity than this explicit loop. Is there a solution with the same complexity as the loop above?

maciekgajewski: The first solution looks very nice,m although, indeed, int will iterate over the collection twice.

Why do you think that the second one would be worse?

florian-kloeck: Because you don't stop iterating after finding the stop value

PR #74, opened 2016-11-22
homework linked list

t-mb:


course4/homework-forward-list/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
+
+set(CMAKE_CXX_STANDARD 14)
```
maciekgajewski, line 3: You still have to modify the CXXFLAGS to0 enable warnings (-Wall)

course4/homework-forward-list/forward_list.hh
```diff
@@ -0,0 +1,100 @@
...
+#include <iterator>
+#include <memory>
+
+namespace course {
+
+template<typename T>
+struct forward_list_node
+{
+	forward_list_node(const T& value, std::unique_ptr<forward_list_node<T>> next)
+		: value(value), next(std::move(next)) { }
```
maciekgajewski, line 14: You copy-copy construct ```value``` here, which is OK... unless it isn't!
Extra assignment: try to make the list work for non-copyable types, for instance ```std::unique_ptr```

course4/homework-forward-list/forward_list.hh
```diff
@@ -0,0 +1,100 @@
...
+template<typename T>
+class forward_list_iterator: public std::iterator<std::forward_iterator_tag, T>
+{
+public:
+	forward_list_iterator(forward_list_node<T>* mStart) : mCurrent(mStart) {}
+
+	const T& operator*() const {
+		return mCurrent->value;
+	}
+	T& operator->() const {
```
maciekgajewski, line 29: what about const operator* and non-const operator ->?

course4/homework-forward-list/forward_list.hh
```diff
@@ -0,0 +1,100 @@
...
+};
+
+
+
+template<typename T>
+class forward_list
+{
+public:
+	forward_list() 
+		: mSize(0), mFront(nullptr) {}
```
maciekgajewski, line 58: You can initialize the members in the point of definition; they you wouldn't have to implement the constructor at all.

course4/homework-forward-list/forward_list.hh
```diff
@@ -0,0 +1,100 @@
...
+
+namespace course {
+
+template<typename T>
+struct forward_list_node
+{
+	forward_list_node(const T& value, std::unique_ptr<forward_list_node<T>> next)
+		: value(value), next(std::move(next)) { }
+
+	const T value;
```
maciekgajewski, line 16: Inconsistent naming: what not mValue?

course4/homework-forward-list/forward_list.hh
```diff
@@ -0,0 +1,100 @@
...
+	forward_list() 
+		: mSize(0), mFront(nullptr) {}
+
+	bool empty() const {
+		return mSize == 0;
+	}
+	std::size_t size() const {
+		return mSize;
+	}
+	const T front() const {
```
maciekgajewski, line 66: forward_list::front returns reference, not copy.

course4/homework-forward-list/forward_list.hh
```diff
@@ -0,0 +1,100 @@
...
+		return mFront->value;
+	}
+	void pop_front() {
+		if (mSize == 0) return;
```
maciekgajewski, line 70: This makes the call safe, but breaks the "don't pay for what you're not using" principle: the if would have to be executed for every call, even if you know that the list is not empty.

Besides: is silent failure a good design?

course4/homework-forward-list/forward_list.hh
```diff
@@ -0,0 +1,100 @@
...
+		mFront = std::move(mFront->next);
+		mSize -= 1;
+	}
+	void push_front(const T& value) {
+		mFront = std::make_unique<forward_list_node<T>>(value, std::move(mFront));
+		mSize += 1;
+	}
+	void push_front(T&& value) {
+		mFront = std::make_unique<forward_list_node<T>>(std::move(value), std::move(mFront));
```
maciekgajewski, line 79: Test it for unnique_ptr, I expect it won't work as expected.

course4/homework-forward-list/forward_list.hh
```diff
@@ -0,0 +1,100 @@
...
+		mSize += 1;
+	}
+
+	forward_list_iterator<T> begin() {
+		return forward_list_iterator<T>(mFront.get());
+	}
+	forward_list_iterator<T> end() {
+		return forward_list_iterator<T>(nullptr);
+	}
+	const forward_list_iterator<T> begin() const {
```
maciekgajewski, line 89: You still can use this function to initialize non-const iterator, and thus modify the container.

maciekgajewski (commented):
Nice, clean and simple; I like it.

PR #75, opened 2016-12-03
matching engine implemention with price-time priority

florian-kloeck:


course4/homework-matching-engine/order_book.hh
```diff
@@ -0,0 +1,183 @@
...
+#include <list>
+#include <map>
+#include <array>
+#include <algorithm>
+
+namespace course {
+
+void print_invalid_order_error();
+
+struct Side
```
maciekgajewski: Please, use enum instead.

course4/homework-matching-engine/order_book.hh
```diff
@@ -0,0 +1,183 @@
...
+{
+	static const int buy = 0;
+	static const int sell = 1;
+};
+
+template<typename PriceType, typename VolumeType>
```
maciekgajewski: What's the point of making this a template?

course4/homework-matching-engine/order_book.hh
```diff
@@ -0,0 +1,183 @@
...
+struct Order
+{
+	std::string order_id;
+	PriceType price;
+	VolumeType volume;
+	int side;
+};
+
+template<typename PriceType>
+class PriceCompareFunctor
```
maciekgajewski: You could've use ```std::less``` and ```std::greater```, instead of custom functor

course4/homework-matching-engine/order_book.hh
```diff
@@ -0,0 +1,183 @@
...
+struct Side
+{
+	static const int buy = 0;
+	static const int sell = 1;
+};
+
+template<typename PriceType, typename VolumeType>
+struct Order
+{
+	std::string order_id;
```
maciekgajewski: Isn't Order ID an integer?

maciekgajewski (commented):
Some unnecessary complexity, unneeded templates.
Also the data is stored in sub-optimal way, but this will be visible in performance showdown, if we ever have one.

PR #81, opened 2017-02-06
My first assignment

anilsukul:
First assignment for C++ course.


maciekgajewski (commented):
Absolutely correct solution ni C! We should use it at some point to compare C vs C++

PR #82, opened 2017-02-07
Extra homework

vargad:
Well, it's a bit ugly... but can be used to compare to Anil's C solution.


maciekgajewski (changes requested):
Don't piggy-back the next assignment on top of the previous one. Please, deliver every one as a separate pullrequest

PR #83, opened 2017-02-07
Extra homework

vargad:
Well, it's a bit ugly... but can be used to compare to Anil's C solution.


maciekgajewski (commented):
Absolutely correct. Nothing ugly here, except for indentation maybe :)

vargad:
I feel bad changing argv. Also because of that the C world leaks into the nice C++ code.

PR #86, opened 2017-02-08
Lazyolddog

ktii:
homework1 pull request Konsta

### DIFF
--- a/course5/homework1/homework1.cc
+++ b/course5/homework1/homework1.cc
@@ -1,7 +1,7 @@
 #include <iostream>
 
 int main(int argc, char** argv) {
-  for (size_t i = 0; i < argc; i++) {
+  for (int i = 0; i < argc; i++) {
     std::cout << i << std::endl;
   }
   return 0;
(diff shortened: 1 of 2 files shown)


course5/homework1/homework1.cc
```diff
@@ -1 +1,8 @@
-// your code goes here
+#include <iostream>
+
+int main(int argc, char** argv) {
+  for (size_t i = 0; i < argc; i++) {
```
maciekgajewski: I expect that with -Wall, you should have a compilation warning here

maciekgajewski (commented):
Good! But I suspect compilation warnings. Aim for a clean, warning-free compilation.

course5/homework1/homework1.cc
ktii: You are right. size_t to int does the trick. cmake/make did not yield a warning. Shouldn't the following make it display warnings: set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -std=c++14") ?

PR #87, opened 2017-02-08
Homework 1

starostinak:


course5/homework1/homework1_hell.cc
```diff
@@ -0,0 +1,43 @@
...
+#include <functional>
+using namespace std;
+
+const int MAX_NUM_ARGS = 445;
+
+template <int I>
+void print(void)
+{
+    print<I - 1>();
+    std::cout << I << " " << std::endl;
```
vargad: Space is not needed here. Very fancy code, nightmare to godbolt :)

course5/homework1/homework1_v2.cc
```diff
@@ -0,0 +1,13 @@
+#include <iostream>
+#include <algorithm>
+#include <cstring>
+
+int main(int argc, char** argv)
+{
+    std::sort(argv, argv + argc,
+            [](const char* s1, const char *s2){ return strcmp(s1, s2) < 0; });
```
maciekgajewski: Use ```std::strcmp```. The unqualified version is a C-call, which is accessible only by accident.

course5/homework1/homework1_v2.cc
```diff
@@ -0,0 +1,13 @@
+#include <iostream>
+#include <algorithm>
+#include <cstring>
+
+int main(int argc, char** argv)
+{
+    std::sort(argv, argv + argc,
```
maciekgajewski: Have you looked at the output from the program?

course5/homework1/homework1_hell.cc
```diff
@@ -0,0 +1,43 @@
...
+}
+
+template <>
+void print<0>(void)
+{
+    std::cout << "0" << std::endl;
+}
+
+std::function<void(void)> printers[MAX_NUM_ARGS];
```
maciekgajewski: You could've dress it in a class, instead of using global variable.

maciekgajewski: Raw function pointer could shave few bytes from the data structure

course5/homework1/homework1_hell.cc
```diff
@@ -0,0 +1,43 @@
...
+template<>
+void definePrinters<0>()
+{
+    printers[0] = &print<0>;
+}
+
+int main(int argc, char **argv)
+{
+    definePrinters<MAX_NUM_ARGS>();
+    printers[argc - 1]();
```
maciekgajewski: You should fall-back to runtime iteration for argcs > MAX_NUM_ARGS

course5/homework1/homework1_hell.cc
```diff
@@ -0,0 +1,43 @@
...
+    return 0;
+}
```
maciekgajewski: Total overkill, but with a potential. To justify it, you should provide a benchmark that proves it's faster than a for loop for some values of argc.

maciekgajewski (commented):
Interesting, but needs work

PR #91, opened 2017-02-10
homework2.1

jfdsmit:


course5/homework2.1/carmack.cc
```diff
@@ -0,0 +1,34 @@
...
+#include <iostream>
+#include <cstdlib>
+
+const float ONE_HALF = 0.5;
+const float THREE_HALFS = 1.5;
+const int FAST_INVERSE_SQRT_MAGIC = 0x5f3759df;
+
+float fastInvSqrt(float input)
+{
+        unsigned long temporaryIntegerForm;
```
maciekgajewski: It is essential that the integer is 32-bits. To achieve portable result, use ```std::uint32_t```

course5/homework2.1/carmack.cc
```diff
@@ -0,0 +1,34 @@
...
+float fastSqrt(float input)
+{
+        return 1 / fastInvSqrt(input);
+}
+
+int main(int argc, char** argv)
+{
+        for (auto i = 1; i < argc; i++)
+        {
+                auto input = std::atof(argv[i]);
```
maciekgajewski: Use ```std::stof``` instead. ```atof``` is na old, C function, and it reports error by return 0.0. stof throws exception, and hence is more explicit and safer.

maciekgajewski (commented):
Very good!

course5/homework2.1/carmack.cc
jfdsmit: Ah, that's the difference, thanks :)

course5/homework2.1/carmack.cc
jfdsmit: Very right, good point

PR #93, opened 2017-02-11
Sqrt

meszenab:
Carmack's algorithm for square root


course5/homework2/carmack/carmack.cc
```diff
@@ -0,0 +1,39 @@
+#include <iostream>
+#include <math.h>
```
maciekgajewski, line 2: Prefer C++ libraruies (```<cmath>, <ctime>```) over the C ones

course5/homework2/carmack/carmack.cc
```diff
@@ -0,0 +1,39 @@
...
+#include <time.h>
+#include <string>
+
+// Implementing fast-sqrt algortihm. However, for me, sqrt from the math library seems much faster.
+
+float inv_sqrt( float x )
+{
+    int i;
```
maciekgajewski, line 10: It is essential that the integer is 32 bit, therefore it is better to use one of the explicitly-sized typedefs (```std::int32_t```)

course5/homework2/carmack/carmack.cc
```diff
@@ -0,0 +1,39 @@
...
+{
+    float x = std::stof(argv[1]);
+    clock_t t;
+    t = clock();
+    std::cout << "Carmack_sqrt of " << argv[1] << "=" << 1/inv_sqrt(x) << std::endl;
+    t = clock() - t;
+    std::cout << "Carmack_time" << "=" << t << std::endl;
+    t = clock();
+    std::cout << "math_sqrt of"  << "=" << sqrt(x) << std::endl;
+    t = clock() - t;
```
maciekgajewski, line 37: It's a good idea to benchmark these two implementations against each other, but not this way.
In this case the time of writing to output will overshadow the time of calculating the square root.
I'll show you how to do it properly one day.

maciekgajewski (commented):
Good

PR #94, opened 2017-02-11
Sqrt newton

meszenab:
Newton's method for sqrt

### DIFF
--- a/course5/homework2/newton/newton.cc
+++ b/course5/homework2/newton/newton.cc
@@ -3,30 +3,27 @@
 
 // Newton method of sqrt. The stack frame was 48 for Debug mode and 64 for Release mode.
 
-bool* start = nullptr;
-bool* end = nullptr;
-
-float function( float y, int it, float& rx, int& ri)
+float newton_sqrt( float y, int it, float x, int i, bool** start_address, bool** end_address)
 {
-    if( ri == 0)
+    if( i == 0)
     {
         bool a;
-        start = &a;
+        *start_address = &a;
     }
-    else if( ri == it )
+    else if( i == it )
     {
         bool a;
-        end = &a;
+        *end_address = &a;
     }
-
-    while( ri < it)
+    
+    if( i < it )
     {
-        ri++;
-        rx -= ( rx * rx - y )/(2 * rx);
-        function( y, it, rx, ri );
+        i++;
+        x -= ( x * x - y )/(2 * x);
+        x = newton_sqrt( y, it, x, i, start_address, end_address );
     }
-
-    return rx;
+    
+    return x;
 
 }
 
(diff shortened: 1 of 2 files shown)


course5/homework2/newton/newton.cc
```diff
@@ -0,0 +1,59 @@
+#include <iostream>
+#include <string>
+#include <time.h>
+
+// Newton method of sqrt. The stack frame was 48 for Debug mode and 64 for Release mode.
+
+int i = 0;
```
maciekgajewski: Didn't I explicitly asked in the alignements not to use global variables?
Maybe I forgot...
Anyway, try again, without global variables!

course5/homework2/newton/newton.cc
```diff
@@ -0,0 +1,59 @@
...
+    if (size > 0)
+    {
+        std::cout << "stack grows downwards" << std::endl;
+    }
+    else if (size < 0)
+    {
+        std::cout << "stack grows upwards" << std::endl;
+    }
+
+    std::cout << "calculation time: " << t << std::endl;
```
maciekgajewski: It isn't calculation time; it's "calculation and printing time"

course5/homework2/newton/newton.cc
```diff
@@ -0,0 +1,59 @@
...
+    const float y = std::stof(argv[1]); // input number
+    x = std::stof(argv[2]); //initial guess
+    const int it = std::stoi(argv[3]); // number of iterations
+
+    clock_t t = clock();
+    std::cout << "sqrt of " << argv[1] << " is " << function(y, it) << std::endl;
+    t = clock() - t;
+
+    std::cout << "depth" << "=" << i + 1 << std::endl;
+    int size = (start - end)/i;
```
maciekgajewski: So, any conclusions? Hoe big is the stack frame? What is the impact of the build type on the stack frame size?

maciekgajewski (commented):
Try without using global variables.
Also, I don't think the number of iterations should be part of the interface.

maciekgajewski (changes requested):
Try without using global variables.
Also, I don't think the number of iterations should be part of the interface.

course5/homework2/newton/newton.cc
meszenab: The stack frame was 48 for debug mode and bigger (64) in release mode

course5/homework2/newton/newton.cc
```diff
@@ -0,0 +1,58 @@
+#include <iostream>
+#include <string>
+
+// Newton method of sqrt. The stack frame was 48 for Debug mode and 64 for Release mode.
+
+bool* start = nullptr;
```
maciekgajewski: These are still global. And you don't need bool. To store pure, untyped address just use void*

course5/homework2/newton/newton.cc
```diff
@@ -0,0 +1,58 @@
...
+bool* end = nullptr;
+
+float function( float y, int it, float& rx, int& ri)
```
maciekgajewski: "function" is not the best name for a function calculating square root. Make your API such so you'd enjoy using it.

course5/homework2/newton/newton.cc
```diff
@@ -0,0 +1,58 @@
...
+int main(int argc, char** argv)
+{
+    const float y = std::stof(argv[1]); // input number
+    float x = std::stof(argv[2]); //initial guess
+    const int it = 1000;
+
+    int i = 0;
+
+    int& ri = i;
+    float& rx = x;
```
maciekgajewski: you don't need these

maciekgajewski (changes requested):
You still have global variables in there

PR #95, opened 2017-02-12
Homework 1 extra, round 2

sophiadavis:


course5/homework1.1/homework1.1.cc
```diff
@@ -0,0 +1,20 @@
...
+int main(int argc, char** argv)
+{
+    std::list<std::string> arg_list;
+
+    for (int i = 0; i < argc; i++)
+    {
+        arg_list.push_back(argv[i]);
+    }
+
+    arg_list.pop_front(); // remove binary name
```
maciekgajewski: If you don't need the first element, why did you put it in the container?

maciekgajewski (commented):
Good. The back-and-forth with argv[0] is unneeded, but otherwise OK

maciekgajewski (commented):
That's better. "Don't pay for what you're not using"

PR #96, opened 2017-02-12
Josef Homework 1 answers

josefschneider:


course5/homework1/homework1.cc
```diff
@@ -1 +1,9 @@
-// your code goes here
+#include <iostream>
+
+int main(int argc, char** argv)
+{
+	for (int counter = 0; counter < argc; counter += 1)
```
maciekgajewski: ++ rather than +=1

course5/homework1.1/homework1.1.cc
```diff
@@ -0,0 +1,17 @@
+#include <iostream>
+#include <set>
+
+int main(int argc, char** argv)
+{
+	std::set<std::string> words;
```
maciekgajewski: Using a b-tree to sort elements is a good strategy, but:
* You store std::string, each of them may require heap allocation - a very costly operation. And thy would only store copies of the strings that are already in memory; wasteful!
* By it's nature, the tree allocates every node on heap. This will be costly, especially that you know the number of elements up front. Try to write an allocator that will serve memory from pre-allcoated array

maciekgajewski (changes requested):
Using set is a valid strategy, but using heap is wasteful. A challenge: keep on using set, but let the program make at most one heap allocation!

course5/homework1.1/homework1.1.cc
vargad: Also consider std::multiset, as you can get the same word multiple times, this way the duplicates get lost.

PR #97, opened 2017-02-12
Homework 2, part 1

sophiadavis:


course5/homework02/homework0201.cc
```diff
@@ -0,0 +1,50 @@
...
+***/
+
+float carmack_sqrt(float input)
+{
+    long* fl_to_int = reinterpret_cast<long*>(&input);
+
+    long shifted_rt = *fl_to_int >> 1;
+    long from_magic = 0x5f3759df - shifted_rt;
+
+    float* int_to_fl = reinterpret_cast<float*>(&from_magic);
```
sophiadavis: Class notes said "Very dangerous!" -- is there another way to do it? (I'll check out some other responses)

maciekgajewski: No, there isn't! This is the "magic" of Carmack's square root; it uses low-level, hardware specific operations, relies on specific implementation of floating-point on the target machine.

course5/homework02/homework0201.cc
```diff
@@ -0,0 +1,50 @@
...
+Implement Carmack’s fast square root. Compare the values it produces with the output of std::sqrt for a range of values, draw conclusions.
+
+Carmack’s fast square root, from Wikipedia:
+an algorithm that estimates 1/sqrt(x), the reciprocal (or multiplicative inverse) of the square root of a 32-bit floating-point number x in IEEE 754 floating-point format.
+The algorithm accepts a 32-bit floating-point number as the input and stores a halved value for later use. Then, treating the bits representing the floating-point number as a 32-bit integer, a logical shift right by one bit is performed and the result subtracted from the magic number 0x5f3759df. This is the first approximation of the inverse square root of the input. Treating the bits again as a floating-point number, it runs one iteration of Newton's method, yielding a more precise approximation.
+***/
+
+float carmack_sqrt(float input)
+{
+    long* fl_to_int = reinterpret_cast<long*>(&input);
```
maciekgajewski: Use size-guaranteed integer type; ```std::int32_t```

course5/homework02/homework0201.cc
```diff
@@ -0,0 +1,50 @@
...
+
+int main(int argc, char** argv)
+{
+    float inputs[23] = {
+        FLT_MIN, 0.000234, 0.001, 0.087, 0.5,
+        5, 13, 19, 52, 78, 9283657, FLT_MAX,
+        0, 1, 4, 9, 16, 25, 36, 49, 400, 1024
+    };
+
+    float carmack_result;
```
maciekgajewski: You don;t have to declare the variables above the loop; you can do it inside, in the place you use them.
Declaring variable of simple type costs nothing; therefore declaring it in the inner scope has 0 cost, while it results in safer and cleaner code.

course5/homework02/homework0201.cc
```diff
@@ -0,0 +1,50 @@
...
+{
+    long* fl_to_int = reinterpret_cast<long*>(&input);
+
+    long shifted_rt = *fl_to_int >> 1;
+    long from_magic = 0x5f3759df - shifted_rt;
+
+    float* int_to_fl = reinterpret_cast<float*>(&from_magic);
+
+    float result = 1 / *int_to_fl;
+    result = result - (( std::pow(result, 2) - input) / (2 * result));
```
maciekgajewski: ```std::pow``` is a complex function, allowing for fractional exponents. Using it here is an overkill, and it's much more expensive than simple```result*result```

maciekgajewski (commented):
Very good!

course5/homework02/homework0201.cc
sophiadavis: I thought I would get a compiler error about `redefinition of carmack_result`. But I don't. 

If I try to include two definitions of the input array, though, I do get that error. What's the difference?

course5/homework02/homework0201.cc
```diff
@@ -13,15 +13,15 @@ The algorithm accepts a 32-bit floating-point number as the input and stores a h
 
 float carmack_sqrt(float input)
 {
-    long* fl_to_int = reinterpret_cast<long*>(&input);
+    int32_t* fl_to_int = reinterpret_cast<int32_t*>(&input);
```
maciekgajewski: These are not really a c++ types, but rater alaises provided by the library.
The proper way of using these is to a) ```#include <cstdint>``` and b) use qualified version: ```std::int32_t```

maciekgajewski (commented):
Good

course5/homework02/homework0201.cc
vargad: I'm not sure what you mean exactly, but declaring a variable in the loop (or in any block ```{ ... }```) creates a variable that is only accessible in that loop (block), it doesn't give you redefinition error because it doesn't mean to create a new variable each iteration.

PR #98, opened 2017-02-12
Homework 2, part 2

sophiadavis:


course5/homework02/stack_exploration.txt
```diff
@@ -0,0 +1,66 @@
...
+Frames remaining  : 2
+  frame start addr: 0x7fff52e67c5c
+Frames remaining  : 1
+  frame start addr: 0x7fff52e67bdc
+Frames remaining  : 0
+  frame start addr: 0x7fff52e67b5c
+16: 4 vs 4
+Initial stack addr: 0x7fff52e680fc
+Last stack addr   : 0x7fff52e67b58
+Average frame size: 32
```
sophiadavis, line 38: wtf? If I subtract 0x7fff52e680fc - 0x7fff52e67b58 and divide by 11, I get 131.3
Is there some weird type casting going on?

course5/homework02/stack_exploration.txt
```diff
@@ -0,0 +1,66 @@
...
+Frames remaining  : 2
+  frame start addr: 0x7fff59e33da4
+Frames remaining  : 1
+  frame start addr: 0x7fff59e33d44
+Frames remaining  : 0
+  frame start addr: 0x7fff59e33ce4
+16: 4 vs 4
+Initial stack addr: 0x7fff59e3410c
+Last stack addr   : 0x7fff59e33cf0
+Average frame size: 23
```
sophiadavis, line 66: ditto -- weird behavior here.

course5/homework02/homework0202.cc
```diff
@@ -0,0 +1,59 @@
...
+}
+
+int main(int argc, char** argv)
+{
+    float input = 16;
+    float guess = 90;
+    int num_iters = 10;
+
+    float* last_on_stack = nullptr;
+    float** addr_of_last_on_stack = &last_on_stack;
```
sophiadavis: ** oh the joy.

course5/homework02/homework0202.cc
```diff
@@ -0,0 +1,59 @@
...
+f(x) = x^2
+f'(x) = 2x
+x_{n+1} = x_{n} - (x_{n}^2 - x_{n}) / (2 * x_{n})
+***/
+
+float newton_sqrt(float input, float result, int num_iters, float** addr_of_last_on_stack)
+{
+    float arbitrarily_reassigned_input = input;
+    std::cout << "Frames remaining  : " << num_iters << std::endl;
+    std::cout << "  frame start addr: " << &arbitrarily_reassigned_input << std::endl;
```
maciekgajewski: why not simply take the address of ```input```?

course5/homework02/homework0202.cc
```diff
@@ -0,0 +1,59 @@
...
+So for square roots:
+f(x) = x^2
+f'(x) = 2x
+x_{n+1} = x_{n} - (x_{n}^2 - x_{n}) / (2 * x_{n})
+***/
+
+float newton_sqrt(float input, float result, int num_iters, float** addr_of_last_on_stack)
+{
+    float arbitrarily_reassigned_input = input;
+    std::cout << "Frames remaining  : " << num_iters << std::endl;
```
maciekgajewski: The debug printing is useful during development, but remove it from the final solution.
You wouldn't like to have a library that prints something from a square root function, do you?

course5/homework02/homework0202.cc
```diff
@@ -0,0 +1,59 @@
...
+    std::cout << "  frame start addr: " << &arbitrarily_reassigned_input << std::endl;
+    if (num_iters > 0) {
+        result = result - (( std::pow(result, 2) - input) / (2 * result));
+        return newton_sqrt(arbitrarily_reassigned_input, result, num_iters - 1, addr_of_last_on_stack);
+    }
+
+    float last_arbitrary_local_variable = 0;
```
maciekgajewski: You just inflate the stack frame with the extra variables, introducing a Heisenberg effect. Do you really need two addresses from every frame?

course5/homework02/homework0202.cc
maciekgajewski: Haha! Exactly. Don't like pointers? Rightly so. Use references instead!

course5/homework02/homework0202.cc
```diff
@@ -0,0 +1,59 @@
...
+
+    float* last_on_stack = nullptr;
+    float** addr_of_last_on_stack = &last_on_stack;
+
+    float newton_result = newton_sqrt(input, guess, num_iters, addr_of_last_on_stack);
+
+    float avg_size = (&input - last_on_stack) / (num_iters + 1);
+
+    std::cout << input << ": " << newton_result << " vs " << std::sqrt(input) << std::endl;
+    std::cout << "Initial stack addr: " << &input << std::endl;
```
maciekgajewski: This is an address from the stack frame of ```main```. I believe that you'd get more accurate results by comparing apples to apples; probing the address from the same point in stack from the same function.

course5/homework02/homework0202.cc
```diff
@@ -0,0 +1,59 @@
...
+x_{n+1} = x_{n} - (x_{n}^2 - x_{n}) / (2 * x_{n})
+***/
+
+float newton_sqrt(float input, float result, int num_iters, float** addr_of_last_on_stack)
+{
+    float arbitrarily_reassigned_input = input;
+    std::cout << "Frames remaining  : " << num_iters << std::endl;
+    std::cout << "  frame start addr: " << &arbitrarily_reassigned_input << std::endl;
+    if (num_iters > 0) {
+        result = result - (( std::pow(result, 2) - input) / (2 * result));
```
maciekgajewski: Prefer ```x*x``` over ```std::pow(x, 2)```

course5/homework02/homework0202.cc
```diff
@@ -0,0 +1,59 @@
...
+    float last_arbitrary_local_variable = 0;
+    *addr_of_last_on_stack = &last_arbitrary_local_variable;
+    return result + last_arbitrary_local_variable;
+}
+
+int main(int argc, char** argv)
+{
+    float input = 16;
+    float guess = 90;
+    int num_iters = 10;
```
maciekgajewski: Should this really be arbitrary?

course5/homework02/stack_exploration.txt
maciekgajewski: No, it's all fine. This is how pointer arithmetic works. Subtraction gives you the difference between two addresses in object sizes; in this case ```sizeof(float)```, 4-bytes.

maciekgajewski (changes requested):
Measurement inaccurate, polluted by unneeded variables.

course5/homework02/stack_exploration.txt
sophiadavis: huh, ok.

course5/homework02/homework0202.cc
sophiadavis: Oh yeah -- we're actually copying the parameter into a local variable in the new stack frame. I didn't think about it as a local variable because it wasn't initialized in the { body } itself.

course5/homework02/homework0202.cc
sophiadavis: Ok, if I cut out all the extra variables that I thought I needed, and just compare the addresses of input at each level of the recursion, then I get 96 bytes per stack frame in the debug build and 80 bytes in the release build.

sophiadavis: But then I still don't see why the program was printing `Average frame size: 32`.

course5/homework02/homework0202.cc
```diff
@@ -0,0 +1,50 @@
...
+
+So for square roots:
+f(x) = x^2
+f'(x) = 2x
+x_{n+1} = x_{n} - (x_{n}^2 - x_{n}) / (2 * x_{n})
+***/
+
+float newton_sqrt(float input, float result)
+{
+    if (result * result - input > .00001) {
```
sophiadavis: I guess this is slightly less arbitrary than just setting the number of iterations beforehand (probably I'm missing something obvious).

But this segfaults if I set the precision too high. If I use `> .000001`, then the segfault occurs trying to calculate the sqrt of 19 (which is 4.3589, according to the output using .00001 as a threshhold). Something to do with floating point precision?

maciekgajewski: You forgot to ABS the diff, This will cause issues if the approximation approaches solution from the wrong side.

Tip: Use named constants for constants

Tip 2: Use scientific notation for better readability


```
const float DELTA = 1E-5;
if (std::abs(result*result - input) > DELTA)
...
```

maciekgajewski (commented):
Cleaner

PR #99, opened 2017-02-13
Homework week2, part1

anilsukul:
Homework week2, part1


course5/week2/ex02.cc
```diff
@@ -0,0 +1,89 @@
...
+
+// ARS - 20170213
+// Cormacks algo from https://en.wikipedia.org/wiki/Fast_inverse_square_root
+// With slight modifications to C++-ifyusing dynamic c
+
+float Q_sqrt(float);
+
+float Q_sqrt( float number )
+{
+	long i;
```
maciekgajewski, line 13: It is important that sizeof(long) == sizeof(float).
1. Use ```std::int32_t``` instead of ```long```
2. Use ```static_assert``` to check the condition at compile time

course5/week2/ex02.cc
```diff
@@ -0,0 +1,89 @@
...
+	y  = * reinterpret_cast<float *> (&i);
+	y  = y * ( threehalfs - ( x2 * y * y ) );   // 1st iteration
+
+	return number*y;
+}
+
+void TimeIt(int max_nr)
+{
+    // OBSERVATIONS: Hack seems to be much slower than std:sqrt???
+    // Iterating over e.g. 1E06 numbers, gives 50ms for Cormacks, vs 7 for std::sqrt. Dos the latter use native assembly sqrt??
```
maciekgajewski, line 33: Would native assembly be faster?
Have you compiled it in debug or in release mode?

course5/week2/ex02.cc
```diff
@@ -0,0 +1,89 @@
...
+
+
+int main(int argc, char** argv)
+{
+    if (argc != 2)
+    {
+		std::cout << "Usage: %s <max_number>" << std::endl;
+		exit(0);
+    }
+    auto max_nr = std::atoi(argv[1]);
```
maciekgajewski, line 83: Use ```std::stoi```

maciekgajewski (commented):
Good.

Timing is tricky; we'll discuss it one day.
Also, you've tested it only fore integer values. More interesting fenomena can be observed for more extreme floating-point numbers.

anilsukul:
Week2, part-2

PR #102, opened 2017-02-16
Homework 2.1

vargad:


course5/homework2.1/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -std=c++14")
+set (CMAKE_CXX_STANDARD 14)
```
maciekgajewski: If you use this, you don't need the ```-std=c++14``` above

course5/homework2.1/homework2.1.cc
```diff
@@ -0,0 +1,31 @@
+#include <iostream>
+#include <cmath>
+#include <cstring>
+
+float fast_inv_sqrt(float number)
+{
+    const float threehalfs = 1.5f;
+    int32_t i;
```
maciekgajewski: Use std::int32_t

maciekgajewski (commented):
Good

PR #103, opened 2017-02-16
Homework 2.2

vargad:


course5/homework2.2/homework2.2.cc
```diff
@@ -0,0 +1,46 @@
...
+};
+
+// compiler: gcc-4.9, linux AMD64 ABI
+// In debug build, stack frame size is 64 byte, compiler generated a recursive function call.
+// In release build, compiler optimized out the recursive function call.
+// The stack grows from the higher address to the lower on this ABI.
+float newton_sqrt(float number, float guess, FuncInspectMetadata &metadata)
+{
+    metadata.ptr_to_local.push_back(&guess);
+    if (std::fabs(guess*guess - number) < EPSILON) return guess;
```
maciekgajewski: Use std::abs instead of std::fabs

course5/homework2.2/homework2.2.cc
```diff
@@ -0,0 +1,46 @@
+#include <iostream>
+#include <cmath>
+#include <cstring>
+#include <vector>
+
+static const float EPSILON = 1.0e-8;
```
maciekgajewski: this constant is local to a function, should be defined inside of the function, to not pollute the namespace.

maciekgajewski (commented):
Good

PR #105, opened 2017-02-18
Homework 3

sophiadavis:
Output: 7truefalse0.33hello, world!66

### DIFF
--- a/course5/homework03/homework03.cc
+++ b/course5/homework03/homework03.cc
@@ -28,7 +28,7 @@ namespace printer
 
 	void print(bool input)
 	{
-		if (input == 1)
+		if (input)
 			std::printf("true");
 		else
 			std::printf("false");
--- a/course5/homework03/homework03.cc
+++ b/course5/homework03/homework03.cc
@@ -1,18 +1,68 @@
 #include <cstdio>
+#include <string>
 
-// TODO your code goes here
+/***
+Imagine that you are a young C++ developer, and your manager, for some reason, forbids std::cout. The only I/O function you are allowed to use to produce the output, is std::printf from <cstdio>
+
+You are not happy about it; you are using C++ after all, and you want to take advantage of the type-safety and convenience it offers.
+
+1. Write a library that provides you with “print”, a function you can use on a selection of types: int, bool, string, float, double etc. The function would use std::printf under the covers to print the value.
+2. For floating-point values, the function could take a second, optional parameter, integer, that specifies the precision with which the number is printed.
+***/
+namespace printer
+{
+	/***
+	provides overloaded function to print various types to stdout
+	***/
+	void print(int input)
+	{
+		std::printf("%i", input);
+	}
+
+	void print(double input, int precision = 0)
+	{
+		std::string format_str = "%." + std::to_string(precision) + "f";
+		const char * format_ptr = format_str.c_str();
+		std::printf(format_ptr, input);
+	}
+
+	void print(bool input)
+	{
+		if (input == 1)
+			std::printf("true");
+		else
+			std::printf("false");
+	}
+
+	void print(const char * input)
+	{
+		std::printf("%s", input);
+	}
+
+	void print(std::string input)
+	{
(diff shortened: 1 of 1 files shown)


course5/homework03/homework03.cc
```diff
@@ -1,18 +1,68 @@
...
-// TODO your code goes here
+/***
+Imagine that you are a young C++ developer, and your manager, for some reason, forbids std::cout. The only I/O function you are allowed to use to produce the output, is std::printf from <cstdio>
+
+You are not happy about it; you are using C++ after all, and you want to take advantage of the type-safety and convenience it offers.
+
+1. Write a library that provides you with “print”, a function you can use on a selection of types: int, bool, string, float, double etc. The function would use std::printf under the covers to print the value.
+2. For floating-point values, the function could take a second, optional parameter, integer, that specifies the precision with which the number is printed.
+***/
+namespace printer
```
sophiadavis, line 12: This is what you meant by "write a library", right?

course5/homework03/homework03.cc
```diff
@@ -1,18 +1,68 @@
...
+
+	void print(bool input)
+	{
+		if (input == 1)
+			std::printf("true");
+		else
+			std::printf("false");
+	}
+
+	void print(const char * input)
```
sophiadavis, line 40: At first I just had the signatures with parameters of type `const char` and `bool`, and it took me forever to get this right (c-style strings kept resolving to the bool-version). What I learned from StackOverflow: "'Hello World' is a string literal of type "array of 12 const char" which can be converted to a "pointer to const char" which can in turn be converted to a bool."

Thanks, compiler.

course5/homework03/homework03.cc
maciekgajewski: It was just a way of adding a bit of role-playing to the task. Creating a library is a bit more complex than just using a namespace; but this is a story for another day.

course5/homework03/homework03.cc
```diff
@@ -1,18 +1,68 @@
...
+	void print(double input, int precision = 0)
+	{
+		std::string format_str = "%." + std::to_string(precision) + "f";
+		const char * format_ptr = format_str.c_str();
+		std::printf(format_ptr, input);
+	}
+
+	void print(bool input)
+	{
+		if (input == 1)
```
maciekgajewski: Comparing boolean to integer is ugly.
"if" expects a boolean expression, so you can just say: ```if(input)```
If you need explicit boolean literals, use ```true``` and ```false```.

course5/homework03/homework03.cc
maciekgajewski: Yep :) This is why you need to suffer trough these assignments; now better way to learn all of this.

course5/homework03/homework03.cc
```diff
@@ -1,18 +1,68 @@
...
+	provides overloaded function to print various types to stdout
+	***/
+	void print(int input)
+	{
+		std::printf("%i", input);
+	}
+
+	void print(double input, int precision = 0)
+	{
+		std::string format_str = "%." + std::to_string(precision) + "f";
```
maciekgajewski: Consider ```"%.*f"```

maciekgajewski (commented):
Good!

course5/homework03/homework03.cc
sophiadavis: neat, did not see that in the docs. Thanks!

course5/homework03/homework03.cc
```diff
@@ -1,18 +1,66 @@
...
+{
+	/***
+	provides overloaded function to print various types to stdout
+	***/
+	void print(int input)
+	{
+		std::printf("%i", input);
+	}
+
+	void print(double input, int precision = 0)
```
vargad: Zero as the default value makes print behave in an unexpected way:
print(6.6); // I think it should print 6.6

sophiadavis: Good point. I peeked at your homework and learned about the 'g' flag. :blush: Thanks!

PR #106, opened 2017-02-18
Homework 3.1

vargad:


maciekgajewski (commented):
Good!

course5/homework03/homework03.cc
```diff
@@ -1,6 +1,35 @@
...
+}
+
+void print(double d)
+{
+    std::printf("%lg", d);
+}
+
+void print(double d, unsigned precision)
+{
+    std::printf("%.*lf", precision, d);
```
sophiadavis, line 21: I can't seem to find the `l` in the docs, what does it do?

vargad: Hi,
Sorry, what you mean? I can't see  ```'``` character in this file.

maciekgajewski: You mean ```.*```?
http://en.cppreference.com/w/cpp/io/c/fprintf

> (optional) . followed by integer number or *, or neither that specifies precision of the conversion. In the case when * is used, the precision is specified by an additional argument of type int. If the value of this argument is negative, it is ignored. If neither a number nor * is used, the precision is taken as zero. See the table below for exact effects of precision.

sophiadavis: No, I meant `l`, I just can't type :woman_facepalming:

vargad: I was under the impression that %f is for float, and %lf is for double. Apperently I was wrong, and older compilers even reject %lf and float is always converted to double when passed to printf. Since C99 and C++11 the standard says %f and %lf is the same thing.

PR #108, opened 2017-02-19
Homework 3 extended

vargad:
Make print function is a bit more convenient.

### DIFF
--- a/course5/homework03/homework03.cc
+++ b/course5/homework03/homework03.cc
@@ -1,18 +1,139 @@
 #include <cstdio>
+#include <string>
+#include <cstring>
+#include <cassert>
+#include <tuple>
+#include <cerrno>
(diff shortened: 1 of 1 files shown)


course5/homework03/homework03.cc
```diff
@@ -1,18 +1,139 @@
...
+                return;
+            }
+            if (format[pos] != '{')
+                hadBracket = false;
+        }
+        ++pos;
+    }
+}
+
+#define fprint(format, ...) \
```
maciekgajewski, line 132: This is FORMATX. The necessity to use macro sucks.
Try avoiding macros. Use C++17 if required.

maciekgajewski (changes requested):
Don't repeat FORMATX mistakes

PR #109, opened 2017-02-20
homework03

anilsukul:
homework03. Added the print functions in a library (libprint.a). Functions cuold have been added to the local homework03.cc file as well, but wanted to play around a bit in cmake.

### DIFF
--- a/course5/homework03/CMakeLists.txt
+++ b/course5/homework03/CMakeLists.txt
@@ -1,5 +1,8 @@
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra ")
 set(CMAKE_CXX_STANDARD 14)
 
-add_executable(homework03 homework03.cc)
+add_library(print print.cc)
+target_link_libraries(print)
+add_executable(homework03 homework03.cc )
+target_link_libraries(homework03 print)
 
--- a/course5/homework03/homework03.cc
+++ b/course5/homework03/homework03.cc
@@ -1,6 +1,7 @@
 #include <cstdio>
-
-// TODO your code goes here
+#include <iostream>
+#include <string>
+#include "print.h"
 
 int main()
 {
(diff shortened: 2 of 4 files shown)


course5/homework03/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra ")
 set(CMAKE_CXX_STANDARD 14)
 
-add_executable(homework03 homework03.cc)
+add_library(print print.cc)
+target_link_libraries(print)
```
maciekgajewski, line 5: This is a no-op

course5/homework03/homework03.cc
```diff
@@ -1,6 +1,7 @@
 #include <cstdio>
-
-// TODO your code goes here
+#include <iostream>
+#include <string>
+#include "print.h"
```
maciekgajewski, line 4: It's better to order headers from the most local to the most general. So the std headers should always go at the end.

course5/homework03/print.h
```diff
@@ -0,0 +1,11 @@
+#include <string>
```
maciekgajewski, line 1: Add ```#pragma once``` at the top of the file. We'll discuss the why later, but this is something that should be added to every header file.

course5/homework03/print.h
```diff
@@ -0,0 +1,11 @@
...
+
+void print(int x);
+void print(long l);
+void print(bool b);
+void print(char x);
+void print(float f, int precision=0);
+void print(double d, int precision=0);
+void print(const char* str);
+void print(std::string s);
```
maciekgajewski, line 10: Once you decided to do it the proper way, you could also wrap all the calls in a namespace.

course5/homework03/print.cc
```diff
@@ -0,0 +1,49 @@
...
+
+void print(float f, int precision)
+{
+	std::string formatter = "%."+ std::to_string(precision) + "f\n";
+	std::printf(formatter.c_str(), f);
+}
+
+void print(double d, int precision)
+{
+	std::string formatter = "%."+ std::to_string(precision) + "lf\n";
```
maciekgajewski, line 37: consider ```%.*f```

maciekgajewski (commented):
Very nice!

PR #110, opened 2017-02-21
Homework3

jfdsmit:


course5/homework03/homework03.cc
```diff
@@ -1,18 +1,70 @@
...
+{
+        print(value.c_str());
+}
+
+void print(double value, unsigned int precision=DEFAULT_PRECISION)
+{
+        std::ostringstream stringStream;
+        stringStream << "%.";
+        stringStream << precision;
+        stringStream << 'f';
```
maciekgajewski, line 53: Consider using ```%.*f```

course5/homework03/homework03.cc
```diff
@@ -1,18 +1,70 @@
 #include <cstdio>
+#include <string>
+#include <sstream>
 
-// TODO your code goes here
+const int DEFAULT_PRECISION = 0;
+const char* TRUE_STRING = "true";
+const char* FALSE_STRING = "false";
```
maciekgajewski, line 7: Bit of an overkill in this case.
Get into habit of marking all namespace-level constants as ```static```. We'll discuss why

maciekgajewski (commented):
Good

PR #111, opened 2017-02-22
Update homework03.cc

AJIEIIIA:
implementation

### DIFF
--- a/course5/homework03/homework03.cc
+++ b/course5/homework03/homework03.cc
@@ -1,6 +1,32 @@
+// Example program
+#include <string>
 #include <cstdio>
 
-// TODO your code goes here
+void print(std::string value)
+{
+	std::printf("%s", value.c_str());
+}
+
+void print(const char* value)
+{
+	std::printf(value);   
+}
+
+void print(bool value)
+{
+	std::printf("%s", value ? "true" : "false");
+}
+
+void print(int value)
+{
+	std::printf("%d", value);
+}
+
+void print(float value, uint precision = 0)
+{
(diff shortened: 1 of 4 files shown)


course5/homework03/homework03.cc
```diff
@@ -1,6 +1,27 @@
...
+}
+
+void print(int value)
+{
+	std::printf("%d", value);
+}
+
+void print(float value, uint precision = 0)
+{
+    	std::string format = std::string("%.") + std::to_string(precision).c_str() + std::string("f");
```
maciekgajewski, line 22: Consider ```%.*f```

maciekgajewski (commented):
Good

AJIEIIIA:
Maciek, why does my program coredump at line 80?

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,96 @@
...
+};
+
+int main(int argc, char** argv)
+{
+
+        Stream Cout;
+
+        // why: double free or corruption (top) ?
+        if (argc > 1)
+                Cout = Stream(argv[1]);
```
vargad, line 80: Here a temporary Stream object is created, which opens the file named argv[1]. The object get copied to Cout object, effectivly coping the FILE pointer to Cout. After the if statement the temporary Stream object gets destroyed, so it closes the file making the pointer invalid, while the Cout object still has pointer to it.

maciekgajewski (changes requested):
Please, deliver every homework in a separate pull request

PR #116, opened 2017-02-25
Homework4

meszenab:


course5/homework4/string.cc
```diff
@@ -0,0 +1,139 @@
...
+        const char* last;
+};
+
+class Stream
+{
+public:
+
+        Stream()
+        {
+            throw std::runtime_error("No file or filename provided");
```
maciekgajewski: So why even provide this function in the first place?

course5/homework4/string.cc
```diff
@@ -0,0 +1,139 @@
...
+            this->file = file;
+            if (!this->file)
+            {
+                throw std::runtime_error("File opening failed");
+            }
+        }
+
+        Stream(const Stream&)
+        {
+            throw std::runtime_error("Stream is not copyable");
```
maciekgajewski: It is much much much much better to enforce such limitations at compile-time rather than at runtime.
Try:
```
Stream(const Stream&) = delete;
```

maciekgajewski (changes requested):
Turn run-time errors into compile-time errors

PR #117, opened 2017-02-25
Homework 4

sophiadavis:


course5/homework04/homework04.cc
```diff
@@ -0,0 +1,122 @@
...
+{
+public:
+	// Constructors
+	Stream() // default
+	{
+		throw std::runtime_error("Stream must be constructed with a filename: Stream('filename.txt')");
+	}
+	Stream(const Stream&) // copy
+	{
+		throw std::runtime_error("Stream object cannot be copied");
```
sophiadavis: I decided copy ~~assignment~~ construction should be ok, because the old object ceases to exist ~~(right?)~~ (edit: I just got this confused with the move constructor), but I don't think it's clear what it would mean for a Stream itself to be copied (Does it stream to the same file as the original object? Does it copy the file to "filename-copy" and start streaming to that one?) So I opted for the error 😃

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,122 @@
...
+	{
+		fname = src.fname;
+		fp = src.fp;
+		return *this;
+	}
+
+	// overloaded << operator
+	Stream& operator << (int input)
+	{
+		if (input == EOF)
```
sophiadavis: Docs say `EOF` is "integer constant expression of type `int` and negative value" -- does this mean that someone could hypothetically pass in an int that happens to be equal to EOF, and not get their desired outcome?

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,122 @@
...
+{
+	Stream s("output.txt");
+
+	std::string w = "World!";
+	s << "Hello, " << w << "\n";
+
+	Cout << "2 + 2 = " << 2 + 2 << "\n";
+	s << "hey" << Endl;
+	s << -2 << Endl;
+	std::abort();
```
sophiadavis: My compiler really didn't like `std::quick_exit` -- I kept getting `error: no member named 'quick_exit' in namespace 'std'` (despite the `#include <cstdlib>`). This happened to work, and the internet seemed to say they're more or less equivalent?

vargad: Which compiler? Visual Studio 2015 and gcc 4.9 should have it.

course5/homework04/homework04.cc
vargad: It should be a compile time error, when copy constructor it used.

maciekgajewski: Why do you think that in case of copy assignment, the copied object  ceases to exists?
```
Class obj1;
Class obj2;
obj1 = obj2; // copy-assignment
obj2.Hello()l // still exists!
```

maciekgajewski: This should be a compile-time error.

course5/homework04/homework04.cc
maciekgajewski: no no no. EOF is used as a return value, not as an input value.
This is the old, C-style way of reporting errors - by returning error codes/special values.

course5/homework04/homework04.cc
maciekgajewski: "abort" could work, but in my setup it actually flushed the buffer (due to the way abort and file streams are implemented).
quick_exit is part of C++11, so please ,check your compiler version and compiler flags.

course5/homework04/homework04.cc
maciekgajewski: Why do you think that in case of copy assignment, the copied object  ceases to exists?
```
Class obj1;
Class obj2;
obj1 = obj2; // copy-assignment
obj2.Hello()l // still exists!
```

maciekgajewski: This should, indeed, be a compile-time error.
You can ask the compiler not to generate the copy constructor:
```
Stream(const Stream&) = delete;
```

maciekgajewski: >  I don't think it's clear what it would mean for a Stream itself to be copied
This is exactly the dilemma that library/class author faces. It's up to you how you will implement this.
I'm a strong believer in simplicity, so simply disallowing copy is the best solution, IMO

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,122 @@
...
+	}
+	Stream(Stream&& src) // move
+	{
+		fname = src.fname;
+		fp = src.fp;
+	}
+	Stream(const char* fname)
+	{
+		this->fname = fname;
+		this->fp = std::fopen(fname, "a");
```
maciekgajewski: What if this fails?

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,122 @@
...
+	{
+		std::fclose(this->fp);
+	}
+
+	// copy assignment operator
+	Stream& operator=(const Stream& src)
+	{
+		fname = src.fname;
+		fp = src.fp;
+		return *this;
```
maciekgajewski: ```src``` lives on

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,122 @@
...
+		throw std::runtime_error("Stream must be constructed with a filename: Stream('filename.txt')");
+	}
+	Stream(const Stream&) // copy
+	{
+		throw std::runtime_error("Stream object cannot be copied");
+	}
+	Stream(Stream&& src) // move
+	{
+		fname = src.fname;
+		fp = src.fp;
```
maciekgajewski: ```src``` will cease to exists immediately after the function is over, but its destructor will be called, so fp will be closed.
You need set src.fp to some value that will stop src's destructor from closing it.

This exposes a gap in your testing: once you implemented the move semantic, you should test it. Write a code that uses it.

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,122 @@
...
+
+Stream Cout(stdout);
+int Endl = EOF;
+
+
+int main(int, char**)
+{
+	Stream s("output.txt");
+
+	std::string w = "World!";
```
maciekgajewski: Test copy-construction, move-construction, copy-assignment and move assignment here.
If you do some error checking - test for it as well.

maciekgajewski (changes requested):
- Make copying fail at compile time
- Checkl if fopen succeeded?
- Test all of it

course5/homework04/homework04.cc
sophiadavis: Maybe it's a because I'm on a mac? Mac also makes it kind of hard to check the gcc version.

`man quick_exit` doesn't give me anything, either.

vargad: I think there is only manual for the C library functions, at least that's the case on linux usually.
Which IDE do use? On Mac XCode I suspect has clang, which is perfect, as long as new enough. Looks like on Mac gcc can easily be a symlink to clang. You can check version in command line with ```gcc --verson``` or ```clang```.

maciekgajewski: I'm sure checking the version is as easy as running ```g++ --version```

sophiadavis: Yeah, I've check that, it spits out Apple gibberish: 
```
g++ --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
(whereas on an ubuntu vm, it gives me a clear "g++ (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609").

vargad: Apperently you have clang on your Mac. Which should be fine, as that clang version is pretty good with the standard, but seems to me that clang still does have quick_exit implemented yet.
https://clang.llvm.org/cxx_dr_status.html

course5/homework04/homework04.cc
sophiadavis: Maybe I'm using it wrong then? Because I'm passing it in as an input when I want a newline - I don't understand what you mean by "return value" here.

course5/homework04/homework04.cc
sophiadavis: I wrote some tests, initially like this:
`Stream s2 = std::move(Stream("another.txt"));`
because I wanted an actual anonymous object that wasn't assigned to a variable name.

But the compiler kept spitting out `warning: moving a temporary object prevents copy elision [-Wpessimizing-move]`, and the "another.txt" file would never be written to. That made it sound like the compiler could NOT do the "copy elision" optimization -- but the copy constructor WAS being called. What was going on?

course5/homework04/homework04.cc
```diff
@@ -125,6 +135,19 @@ int main(int, char**)
...
-	std::abort();
+	yolo(std::move(s));
+
+	Stream sToMove("another.txt");
+	Stream s2 = std::move(sToMove); // test move constructor
+	s2 << "\n :) \n" << Endl;
+
+	Stream s3("more.txt");
+	s3 = std::move(sToMove);  // test move assignment
+	s3 << "\n :( \n" << Endl;
```
sophiadavis: I still haven't implemented the "don't close the file if it's already a nullptr" bit, but this isn't segfaulting... I added printf to the destructor, and it says it's closing the file twice. Wut.  Does the `std::move` somehow magically keep both objects in scope? What am I doing wrong?

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,127 @@
...
+	// s2 = s;
+
+	// Stream s2("/etc/passwd"); // test that file-open error checking is working
+	// s2 << "YOLO!" << Endl;
+
+	// Stream sToMove("another2.txt"); // test that move construction doesn't compile
+	// Stream s2 = std::move(sToMove);
+
+	// Stream sToMove("another2.txt"); // test that move assignment doesn't compile
+	// s = std::move(sToMove);
```
sophiadavis: On second thought, I decided to delete the move construction and assignment operators. In these lines, if I then tried to write to `sToMove` after move-assigning to `s`, it caused a segfault (because `sToMove`'s file handle had already been closed), and that felt weird. 

If you could really have anonymous temporary Streams, then move construction/assignment would make sense, but I can't figure out what an 'anonymous temporary Stream' even would be (it *has* to have a filename, so it can't really be anonymous at all...)

maciekgajewski: Once you establish that a object of a class has an unique identity (and we established that Stream certainly has), then you should do your utmost to replace copy assignment/constructor with move equivalents, so that you can move the unique identity between different objects.

Examples:

```
Stream OpenLogFile()
{
   auto path = ReadLogFilePathFromSettings();
   CheckPermisssions(path);
   return Stream(path, Stream::APPEND);
}
// ...
Stream log = 
Stream log = OpenLogFile(); // move construction
// ..
```

another example:
```
std::vector<Stream> outputFiles;
void AddOutputFile(path)
{
   Stream file(path, Stream::WRITE);
   outputFiles.push_back(std::move(file));
}

```

sophiadavis: Ah, I had thought of move only being used from anonymous temporary objects. What you're saying about "transferring unique identity" makes it a lot clearer -- so did what you said in class about moved objects being "zombies". Thanks!

PR #120, opened 2017-02-27
Homework week4

anilsukul:


course5/homework04/Stream.cc
```diff
@@ -0,0 +1,79 @@
+#include "Stream.h"
+#include <cstdio>
+#include <string>
+#include <stdexcept>
+
+Stream::Stream(const char* file)
+{
+	if ((m_fileptr = std::fopen(file, "w")) == nullptr)
+	{
+		throw (std::runtime_error("Cannot open file\n"));
```
maciekgajewski, line 10: 1. No need to add parenthesis around the exception.
2. The endline at the end of the message will only cause problems for someone who wants to handle the error in any other way than just printing it to the console.

course5/homework04/Stream.h
```diff
@@ -0,0 +1,41 @@
...
+};
+
+
+/* 
+Copyable? 
+   A copy constructor for a stream is technically possible, but practically not wise, as
+   you will have 2 objects having the same file descriptor. If the one objects is destructed, the
+   other is stuck with an invalid filepointer. A possibility would be to have the copy contructor not
+   copy the filepointer, but open the file with its own filepointer or dup() the other, but again, 
+   synchronizing might become an issue
```
maciekgajewski, line 33: Agreed. But your implementation is copyable - it has compiler-generated copy constructor.
Vide: http://en.cppreference.com/w/cpp/language/copy_constructor

course5/homework04/Stream.h
```diff
@@ -0,0 +1,41 @@
...
+
+Default constructable?
+   Why not, if the definition e.g. states that it will write to stdout as a default?
+
+Movable
+   Unlike copyable, I think a move contructor makes sense, as it moves all the resources from the src,
```
maciekgajewski, line 39: Agreed as well. But your implementations does not provide move constructor/assignment.

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,47 @@
...
+		std::cout << "Testing Cout" << std::endl; 
+		Cout << "2 + 2 = "<< 2 + 2 << "\n"; 
+		std::cout << "done" << std::endl; 
+	
+		std::cout << "Testing Stream object with flush() to output file file.txt " << std::endl; 
+		Stream f("file.txt");
+		f << "hey" << Endl;
+		std::quick_exit(0);
+		
+	}
```
maciekgajewski, line 41: Please, try and see if copy/move initialisation/assignment works.

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,47 @@
...
+	catch (const std::exception& e)
```
maciekgajewski, line 42: It is a good practice not to catch exception on the top level, but rather to let if flow.
This way you'll not only get the error message, but more meta-data and a core dump, with a stack trace to the point when the exception was thrown.

We'll discuss it in more details.

maciekgajewski (changes requested):
Delete copy constructor; add move constructor/assignment.
Test it.

course5/homework04/Stream.cc
```diff
@@ -0,0 +1,79 @@
...
+Stream& Stream::operator << (const char* s)
+{
+	std::fprintf(m_fileptr, "%s", s);
+	return *this; 
+}
+
+Stream& Stream::operator << (std::string s)
+{
+	std::fprintf(m_fileptr, "%s", s.c_str());
+	if (s == Endl)
```
maciekgajewski, line 48: This works as requested, but:
1. User may accidentally trigger a flush, by passing a string of certain value. Violates the Principle of Least Surprise
2. This runtime check (with complexity O(Endl.length())) is needed every time a string is printed. This violates the Don't Pay For What You Are Not Using. The runtime check could be replaced by compile-time dispatch.

PR #121, opened 2017-03-04
vector homework

meszenab:


course5/homework5/vector.cc
```diff
@@ -0,0 +1,131 @@
+#include "vector.h"
+#include <iostream>
+#include <cmath>
+
+#define Pi 3.14159265
```
maciekgajewski: cmath provides the constant

course5/homework5/vector.cc
```diff
@@ -0,0 +1,131 @@
...
+
+    Vector2D operator*(double r, const Vector2D& V)
+    {
+        Vector2D W(r * V.GetX(), r * V.GetY());
+        return W;
+    }
+
+}
+
+int main(int, char**)
```
maciekgajewski: As per assignment description: please, put main into a separate file

course5/homework5/vector.h
```diff
@@ -0,0 +1,39 @@
...
+        //    Vector2D(Vector2D&&);
+            Vector2D& operator=(const Vector2D&);
+        //    Vector2D&& operator=(Vector2D&& src);
+            double GetX() const;
+            double GetY() const;
+            void SetX(double);
+            void SetY(double);
+            double GetR() const;
+            double GetAngle() const;
+            Vector2D operator+(const Vector2D&);
```
maciekgajewski: Is there a reason why operator+ can't be const?

course5/homework5/vector.h
```diff
@@ -0,0 +1,39 @@
...
+            Vector2D operator-(const Vector2D&);
```
maciekgajewski: Unary minus operator is missing

course5/homework5/vector.h
```diff
@@ -0,0 +1,39 @@
...
+            Vector2D& operator+=(const Vector2D&);
+            Vector2D& operator-=(const Vector2D&);
+            bool operator==(const Vector2D&);
+            bool operator!=(const Vector2D&);
```
maciekgajewski: these should be const too

course5/homework5/vector.cc
```diff
@@ -0,0 +1,131 @@
...
+#define Pi 3.14159265
+
+namespace vector{
+
+    Vector2D::Vector2D(double x, double y) : x(x), y(y) {}
+    Vector2D::Vector2D(const Vector2D& src) : x(src.x), y(src.y)  {}
+    Vector2D& Vector2D::operator=(const Vector2D& src)
+    { 
+        x = src.x;
+        y = src.y;
```
maciekgajewski, line 13: something's wrong with this line

course5/homework5/vector.cc
```diff
@@ -0,0 +1,131 @@
...
+    {
+        this->x = x;
+    }
+
+    void Vector2D::SetY(double y)
+    {
+        this->y = y;
+    }
+
+    double Vector2D::GetR() const { return sqrt(x*x + y*y); }
```
maciekgajewski: no way this compiles!

course5/homework5/vector.cc
```diff
@@ -0,0 +1,131 @@
...
+
+    Vector2D operator "" _x(long double x)
+    {
+        Vector2D X(x, 0);
+        return X;
+    }
+
+    Vector2D operator "" _y(long double y)
+    {
+        Vector2D Y(0, y);
```
maciekgajewski, line 82: broken intentation

course5/homework5/vector.cc
```diff
@@ -0,0 +1,131 @@
...
+        return Y;
+    }
+
+    Vector2D operator "" _phi(long double phi) // user can provide phi in degree
```
maciekgajewski: Why in degree? All functions in the math lib expect input in radians

course5/homework5/vector.cc
```diff
@@ -0,0 +1,131 @@
...
+    Vector2D& Vector2D::operator-=(const Vector2D& src)
+    {
+        x -= src.x;
+        y -= src.y;
+        return *this;
+    }
+
+    bool Vector2D::operator==(const Vector2D& src)
+    {
+        if ( x == src.x && y == src.y )
```
maciekgajewski: the only purpose of the entire 4-line if statement id to convert bool to bool

course5/homework5/vector.cc
maciekgajewski: Please, qualify the library functions: ```std::sqrt```, ```std::sin``` etc

course5/homework5/vector.cc
```diff
@@ -0,0 +1,131 @@
...
+
+    std::cout << "u = " << u << std::endl;
+
+    v += u;
+
+    std::cout << "v = " << v << std::endl;
+
+    std::cout << "v == u: " << (v == u) << std::endl;
+
+    std::cout << "v != u: " << (v != u) << std::endl;
```
maciekgajewski: Many of the methods remain untested, for instance the getters for radial coords.

maciekgajewski (changes requested):
- put main into a separate file
- add unary minus implementation
- expand tests to cover all methods

course5/homework5/vector.cc
meszenab: It does compile and gives the expected result. What make you say that? sqrt vs. std::sqrt?

course5/homework5/vector.h
meszenab: You mean that the method is const, so
Vector2D operator+(const Vector2D&) const; ?

course5/homework5/vector.cc
meszenab: Hm, seems to work for me. Could you elaborate?

course5/homework5/vector.cc
maciekgajewski: You must have missed this part of the lecture:
```sqrt``` is a C function that compiles and works here sort of by accident. But C++ only guarantees ```std::sqrt```. There may be a C+ compiler now, or in the future, that will fail to compile the code.

course5/homework5/vector.cc
maciekgajewski: I looked differently when I was looking a it. Maybe github was tricking me...?

course5/homework5/vector.h
maciekgajewski: yes

PR #122, opened 2017-03-04
Dani homework 5

vargad:


course5/homework05/vector.h
```diff
@@ -0,0 +1,87 @@
...
+class Radian {
+    double mRadian;
+public:
+    explicit Radian(double radian) : mRadian(radian) {}
+    Radian(const Degree& degree);
+
+    double Get() const { return mRadian; }
+};
+
+class Degree {
```
maciekgajewski: Why don't you just define one class, called 'Angle'?

course5/homework05/vector.h
```diff
@@ -0,0 +1,87 @@
...
+{
+public:
+    explicit PolarHelper(double r)
+        : mR(r)
+    {}
+
+    double Get() { return mR; }
+private:
+    double mR;
+};
```
maciekgajewski, line 34: Ans another class called "Length"

course5/homework05/vector.h
```diff
@@ -0,0 +1,87 @@
...
+    double GetR() const { return std::sqrt(mX*mX + mY*mY); }
+    double GetO() const { return std::atan2(mY, mX); }
+
+    void SetX(double x) { mX = x; }
+    void SetY(double y) { mY = y; }
+private:
+    double mX, mY;
+};
+
+Vector2D operator+(PolarHelper r, Radian o);
```
maciekgajewski: I think that operator* makes more sense here than operator+. Just a gut feeling.

course5/homework05/vector.h
```diff
@@ -0,0 +1,87 @@
...
+    Vector2D(double x, double y);
+    Vector2D(Vector2D const &) = default;
+    Vector2D(Vector2D &&) = default;
+    Vector2D& operator=(Vector2D const &) = delete;
+    Vector2D& operator=(Vector2D &&rhs) = default;
+
+    // vector small enough that it likely better to take as value
+    bool operator==(Vector2D rhs);
+    bool operator!=(Vector2D rhs);
+    Vector2D operator-();
```
maciekgajewski, line 48: Unary minus operator is missing

maciekgajewski (changes requested):
Good, but unary minus is missing

course5/homework05/vector.h
vargad: What you mean on unary minus? I think it's right there.

maciekgajewski: ah yes. but it isn't tested.
BTW: you could automate the tests

PR #123, opened 2017-03-04
Homework04

AJIEIIIA:


.gitignore
```diff
@@ -1,5 +1,6 @@
 #$ build dir
 build_
+course5/builds
```
maciekgajewski, line 3: This is one of the reasons I believe it's better to keep your builds outside of your source directory

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,115 @@
...
+        {
+                file = source.file;
+                source.file = nullptr;
+        }
+
+        Stream(const Stream& other) = delete;
+
+	~Stream()
+        {
+                if (file && file != stdout)
```
maciekgajewski: ```stdout``` is not the only special file, there is also ```stdin``` and ```stderr```

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+        Stream& operator = (Stream&& stream)
+        {
+                file = stream.file;
+                stream.file = nullptr;
+                return *this;
+        }
+
+	Stream& operator << (std::string value)
```
maciekgajewski: pass ```std::string``` and other potentially expensive to copy objects by const reference

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,115 @@
...
+        cout << 1.0/3.0;// 0.33
+
+        std::string s = "world!";
+        cout << "hello, "; // hello,
+        cout << s; // world!
+        cout << Stream::endl;
+
+        float f = 6;
+        cout << f; // 6
+        cout << Stream::endl;
```
maciekgajewski, line 111: Extra question: can you fully automate the test? Can you come up with a way to programmatically test if the output was correct?

maciekgajewski (commented):
Very nice!

course5/homework04/homework04.cc
AJIEIIIA: agree. fixed

course5/homework04/homework04.cc
AJIEIIIA: yes, forgot about it. fixed.

course5/homework04/homework04.cc
AJIEIIIA: while implementing this test I found a bug. I haven't implemented print method taking char * as a parameter. apparently compiler casted char * to bool and the output was incorrect (true instead of hello, ). Now I wonder why was compiler choosing bool, not std::string overload?

maciekgajewski: Nasty C-legacy: implicit conversion between numeric types, and between pointers to void* and bool.
It seems that it tales precedence over user-defined conversion, like const char* -> std::string

PR #124, opened 2017-03-05
Homework5

josefschneider:


course5/homework5/main.cc
```diff
@@ -0,0 +1,74 @@
+#include "vector2d.h"
+
+#include <iostream>
+#include <cassert>
+#include <cmath>
+
+const double pi = std::acos(-1);
```
maciekgajewski: M_PI

course5/homework5/vector2d.h
```diff
@@ -0,0 +1,51 @@
...
+struct RadialCoord { double val; };
+struct AngularCoord { double val; };
+
+inline Vector2D operator "" _x(long double x) { return Vector2D(x, 0.0); }
+inline Vector2D operator "" _y(long double y) { return Vector2D(0.0, y); }
+inline RadialCoord operator "" _r(long double r) { return RadialCoord{static_cast<double>(r)}; }
+inline AngularCoord operator "" _t(long double t) { return AngularCoord{static_cast<double>(t)}; }
+
+Vector2D operator+(const RadialCoord& r, const AngularCoord& t);
+Vector2D operator+(const AngularCoord& t, const RadialCoord& r);
```
maciekgajewski: I _feel_ that operator* would be a better fit for this operation

course5/homework5/vector2d.h
```diff
@@ -0,0 +1,51 @@
...
+
+	// Getters for the radial components
+	double GetRad() const;
+	double GetTheta() const;
+
+	bool operator==(const Vector2D& lhs) const;
+	bool operator!=(const Vector2D& lhs) const;
+	Vector2D operator+(const Vector2D& lhs);
+	Vector2D& operator+=(const Vector2D& lhs);
+	Vector2D operator-(const Vector2D& lhs);
```
maciekgajewski, line 28: Unary minus is missing

course5/homework5/vector2d.h
```diff
@@ -0,0 +1,51 @@
+#include <ostream>
```
maciekgajewski, line 3: Add ```#pragma once```

course5/homework5/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
+set(CMAKE_CXX_STANDARD 14)
+
+add_executable(homework5 main.cc vector2d.cc)
```
maciekgajewski: headers should also be added to the list of sources

maciekgajewski (changes requested):
Nice. But unary minus is missing

course5/homework5/vector2d.h
josefschneider: I thought I had implemented it. The unary minus is the -= right?

maciekgajewski: No. Unary minus is this:
```
Vector2D a(3, 4);
assert(-a == Vector2D(-3. -4));
```

josefschneider: Ah yes, of course

course5/homework5/vector2d.h
```diff
@@ -0,0 +1,54 @@
...
+struct AngularCoord { double val; };
+
+inline Vector2D operator "" _x(long double x) { return Vector2D(x, 0.0); }
+inline Vector2D operator "" _y(long double y) { return Vector2D(0.0, y); }
+inline RadialCoord operator "" _r(long double r) { return RadialCoord{static_cast<double>(r)}; }
+inline AngularCoord operator "" _t(long double t) { return AngularCoord{static_cast<double>(t)}; }
+
+Vector2D operator-(const Vector2D& v);
+Vector2D operator*(const RadialCoord& r, const AngularCoord& t);
+Vector2D operator*(const AngularCoord& t, const RadialCoord& r);
```
sophiadavis, line 46: Neat, I couldn't figure this out. And you make it really clear.

course5/homework5/vector2d.h
```diff
@@ -0,0 +1,54 @@
...
+	double y = 0.0;
+};
+
+struct RadialCoord { double val; };
+struct AngularCoord { double val; };
+
+inline Vector2D operator "" _x(long double x) { return Vector2D(x, 0.0); }
+inline Vector2D operator "" _y(long double y) { return Vector2D(0.0, y); }
+inline RadialCoord operator "" _r(long double r) { return RadialCoord{static_cast<double>(r)}; }
+inline AngularCoord operator "" _t(long double t) { return AngularCoord{static_cast<double>(t)}; }
```
sophiadavis, line 42: Why do you mark them explicitly as `inline`?

josefschneider: I believe this is the only useful use of "inline" these days: to make sure that this function in the header is not defined more than once. If you remove the inline the code will generally not compile.

Using inline as an optimisation technique is pointless in the days of modern compilers as they tend to be much better than us at figuring out how to optimise our code

vargad: That's right, generally bad to use inlie for optimalization. Without inline the code compiles just fine, but each time you include the header the compiler compiles the function into the object file (because this is definition, not just a declaration). After compile the linker builds the binary using the object files, but it will fail because multiple object files has the same 'symbol' (function). That is the error message Sophia saw before.

PR #128, opened 2017-03-09
Fei homework04, not finished, stuck at the move assignment...

feicui08:
cannot build because of the error on move assignment. 
expected ',' or '...' before '&&' token.
'src ' was not declared in this scope.

### DIFF
--- a/course5/homework04/homework04.cc
+++ b/course5/homework04/homework04.cc
@@ -1,0 +1,79 @@
+#include <cstdio>
+#include <cstdlib>
+#include <string>
+#include <cstring>
+
+
+struct FlushEndl
+{
+
+};
+
+class Stream
+{
+public:
+    Stream(const std::string &OutFileName)
+    {
+        output_file = std::fopen(OutFileName.c_str(), "w");
+    }
+
+    Stream(const Stream& other) = delete;
+
+    ~Stream()
+    {
+        std::fclose(output_file);
+    }
+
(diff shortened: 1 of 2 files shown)


vargad:
It builds for me, are you sure it uses the newer comiler? You can check by running ```make VERBOSE=1``` in the build directory, the gcc/g++ it runs should be at least gcc 4.9, and it has to pass -std=c++14 flag to the compiler.

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    Stream& operator <<(FlushEndl)
+    {
+        std::fprintf(output_file, "%s", "\n");
+        std::fflush(output_file);
+        return *this;
+    }
+
+    Stream& operator =(Stream &&src)
```
maciekgajewski, line 52: you'd also need a move constructor

course5/homework04/homework04.cc
```diff
@@ -0,0 +1,79 @@
...
+};
+
+
+Stream Cout("/dev/stdout");
+FlushEndl Endl;
+
+
+int main()
+{
+    Stream s("output.txt");
```
maciekgajewski, line 70: test move constructor, test move assignment

maciekgajewski (changes requested):
Add move constructor
Test move constructor and move assignment

PR #131, opened 2017-03-12
Homework6

josefschneider:


course5/homework6/string.cc
```diff
@@ -0,0 +1,106 @@
+
+#include "string.h"
+
+namespace String {
+
+String::String(const String& copy):
+	mContents(std::make_unique<char[]>(mSize)),
+	mSize(copy.mSize)
```
maciekgajewski: I'm surprised that it works. I would expect mSize to be not initialized yet in line 7.
Isn't valgrind complaining here?

course5/homework6/string.cc
```diff
@@ -0,0 +1,106 @@
...
+
+String::String(const char* charString)
+{
+	*this = charString;
+}
+
+String& String::operator=(const char* lhs)
+{
+	// Find the end of the string
+	while (lhs[mSize] != '\0')
```
maciekgajewski: aka std::strlen

course5/homework6/string.cc
```diff
@@ -0,0 +1,106 @@
...
+	mContents = std::move(move.mContents);
+	return *this;
+}
+
+String::String(const char* charString)
+{
+	*this = charString;
+}
+
+String& String::operator=(const char* lhs)
```
maciekgajewski: It should be rhs, not lhs, right?

course5/homework6/string.cc
```diff
@@ -0,0 +1,106 @@
...
+
+String& String::operator+=(const String& lhs)
+{
+	*this = std::move(*this + lhs);
+	return *this;
+}
+
+char& String::operator[](const int idx)
+{
+	if (static_cast<unsigned>(idx) >= mSize)
```
maciekgajewski: Why not simply use the same integer type for mSize and idx?

course5/homework6/string.cc
```diff
@@ -0,0 +1,106 @@
...
+		throw std::out_of_range("string index out of range");
+	return mContents[idx];;
+}
+
+const char& String::operator[](const int idx) const
+{
+	if (static_cast<unsigned>(idx) >= mSize)
+		throw std::out_of_range("string index out of range");
```
maciekgajewski: Q: Is std::string doing range checking in operator[]?

course5/homework6/string.cc
```diff
@@ -0,0 +1,106 @@
...
+	return mContents[idx];;
+}
+
+bool String::operator==(const String& lhs) const
+{
+	if (mSize != lhs.mSize)
```
maciekgajewski: try this:
```assert(string("") == string());```

File: course5/homework6/string.cc [hunk 7]
```diff
@@ -0,0 +1,106 @@
...
+		return false;
+
+	for (size_t idx = 0; idx < mSize - 1; idx++)
```
maciekgajewski: std::strcmp or std::equal

course5/homework6/main.cc
```diff
@@ -0,0 +1,135 @@
...
+#include <string.h>
+#include <cassert>
+
+int main()
+{
+	{
+		// First check if const char* constructor and c_str() method
+		// work as these will be used for tests later on
+		String::String s("abc");
+		assert(::strcmp(s.c_str(), "abc") == 0);
```
maciekgajewski: ```std::strcmp```

course5/homework6/main.cc
```diff
@@ -0,0 +1,135 @@
+
+#include "string.h"
+
+#include <iostream>
+#include <string.h>
```
maciekgajewski: ```#include <cstring>```

maciekgajewski (commented):
Mostly good; there are some corner case when I expect tests to fail.
Have you run it past valgrind?

course5/homework6/string.cc
josefschneider: I forgot to run Valgrind - all fixed now

course5/homework6/string.cc
josefschneider: Yup, silly mistake!

course5/homework6/string.cc
josefschneider: I hadn't realised we were allowed to use cstring methods, this makes it indeed a lot easier!

course5/homework6/string.cc
josefschneider: To me it made sense to have idx as an int as that is the usual type for an index for operator[]. I've changed it to size_t

course5/homework6/string.cc
josefschneider: No it isn't but I thought it might be a good addition. I suppose that seeing as none of these throw I should really add noexcept to a lot of these methods?

course5/homework6/string.cc
josefschneider: Fixed

PR #133, opened 2017-03-13
Homework06

sophiadavis:
🎆  🍾  🎆  🍾  🎆  🍾  🎆  🍾  🎆  🍾  🎆  🍾  🎆  🍾  🎆  🍾  🎆  🍾  🎆  🍾  🎆  🍾  🎆  🍾  🎆  🍾  🎆  🍾  
```
==10752== HEAP SUMMARY:
==10752==     in use at exit: 26,597 bytes in 190 blocks
==10752==   total heap usage: 267 allocs, 77 frees, 32,497 bytes allocated
==10752== 
==10752== LEAK SUMMARY:
==10752==    definitely lost: 0 bytes in 0 blocks
==10752==    indirectly lost: 0 bytes in 0 blocks
==10752==      possibly lost: 2,064 bytes in 1 blocks
==10752==    still reachable: 4,096 bytes in 1 blocks
==10752==         suppressed: 20,437 bytes in 188 blocks
==10752== Rerun with --leak-check=full to see details of leaked memory
==10752== 
==10752== For counts of detected and suppressed errors, rerun with: -v
==10752== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

### DIFF
--- a/course5/homework06/main.cc
+++ b/course5/homework06/main.cc
@@ -34,8 +34,13 @@ void test(const char* test_name, const StrT& s, const char* expected_content, in
     int content_l = std::strlen(expected_content);
     bool strncmp_result = std::strncmp(s.c_str(), expected_content, std::min(s_l, content_l));
 
-    assert(strncmp_result == 0);
     assert(s.length() == expected_length);
+    assert(strncmp_result == 0);
+
+    for (auto i = 0; i < expected_length; i++)
+    {
+        assert(s[i] == expected_content[i]);
+    }
 
     std::cout << "Passed: " << test_name << std::endl;
 }
--- a/course5/homework06/main.cc
+++ b/course5/homework06/main.cc
@@ -29,58 +29,9 @@ int main(int, char**)
     using namespace simple_string;
     String s;
     assert(s.Length() == 0);
-    std::cout << s.Length() << std::endl;
     s.Print();
 
-
-    std::unique_ptr<char[]> p = std::make_unique<char[]>(5);
-    for(int i = 0; i < 4 ; i++)
-    {
-        p[i] = 'y';
-    }
-    for(int i = 0; i < 3 ; i++)
-    {
-        std::cout << p[i] << std::endl;
-    }
-    p[0] = 'y';
-    p[1] = 'o';
-    p[2] = 'l';
-    p[3] = 'o';
-
-    std::cout << *p.get() << std::endl;
-
-    // int y = 0;
-    // int * ptr = &y;
-    // *ptr = 8;
-    //
-    // std::cout << *ptr << std::endl;
-
-    char thechars[] = "chars";
-    std::cout << thechars << std::endl;
-
-    std::cout << std::strlen(thechars) << std::endl;
-    const char * res = "";
-    for (auto i = 0; i < 5; i++)
-    {
-        res += thechars[i];
-    }
-    std::cout << res << "  RES" << std::endl;
-
-    // char thechars2[] = *p;
-    // std::cout << thechars2 << std::endl;
-    std::unique_ptr<int[]> int_p = std::make_unique<int[]>(5);
-    for (auto i = 0; i < 5; i++)
-    {
-        int_p[i] = i;
-    }
-
-    std::cout << int_p[3] << std::endl;
-
-
-    const char* word = "word";
-    std::cout << std::strlen(word) << std::endl;
-
-    String sFromChar = word;
+    String sFromChar = "word";
     sFromChar.Print();
 
     String copyConstructedString = sFromChar;
@@ -90,7 +41,8 @@ int main(int, char**)
     String moveConstructedString = std::move(sFromChar);
     moveConstructedString.Print();
 
-    String assignmentTester = word;
+    String assignmentTester = "yolo";
+    assignmentTester.Print();
     assignmentTester = moveConstructedString;
     assignmentTester.Print();
     moveConstructedString.Print();
@@ -124,10 +76,26 @@ int main(int, char**)
 
     // assignmentTester[5] = 'c'; // should raise runtime error
 
-    s = "bonanza";
-    std::cout << std::strlen("bonanza") << std::endl;
-    std::cout << s.Length() << std::endl;
-    std::cout << s << std::endl;
+    String concatTester1 = "Hello ";
+    String concatTester2 = concatTester1 + "World!";
+    concatTester2.Print();
+    concatTester1.Print();
+
+    String concatTester3 = "World Again!";
+    String concatTester4 = concatTester1 + concatTester3;
+    concatTester4.Print();
+
+    concatTester4 += " And again.";
+    concatTester4.Print();
+
+    String concatTester5 = " And again twice.";
+    concatTester4 += concatTester5;
+    concatTester4.Print();
+
+    // s = "bonanza";
+    // std::cout << std::strlen("bonanza") << std::endl;
+    // std::cout << s.Length() << std::endl;
+    // std::cout << s << std::endl;
     // std::cout << s.CStr() << std::endl;
     // assert(s.CStr() == "bonanza");
 
--- a/course5/homework06/main.cc
+++ b/course5/homework06/main.cc
@@ -47,12 +47,55 @@ int main(int, char**)
     using namespace std::literals;
 
     String defaultS;
-    const int num_test_cases = 1;
-    std::array<const char *, num_test_cases> test_names = { {"default constructor"} };
-    std::array<String, num_test_cases> inputs = { {defaultS} };
-    std::array<const char *, num_test_cases> expected_contents = { {""} };
-    std::array<int, num_test_cases> expected_lengths = { {0} };
+    String sFromChar = "word";
+    String copyConstructed = sFromChar;
 
+    String toBeAssigned = "movetest";
+    String moveConstructed = std::move(toBeAssigned);
+
+    toBeAssigned = "assignment";
+    String copyAssigner = "CopyAssignER";
+    String moveAssigner = "MoveAssignER";
+    String copyAssigned;
+    copyAssigned = copyAssigner;
+    String moveAssigned;
+    moveAssigned = std::move(moveAssigner);
+
+    const int num_test_cases = 7;
+    std::array<const char *, num_test_cases> test_names = {
+        {
+            "default constructor",
+
+            "const char * construction",
+            "copy construction",
+            "move construction",
+
+            "const char * assignment",
+            "copy assignment",
+            "move assignment"
+        }
+    };
+    std::array<String, num_test_cases> inputs = {
+        {
+            defaultS,
+            sFromChar, copyConstructed, moveConstructed,
+            toBeAssigned, copyAssigned, moveAssigned
+        }
+    };
+    std::array<const char *, num_test_cases> expected_contents = {
+        {
+            "",
+            "word", "word", "movetest",
+            "assignment", "CopyAssignER", "MoveAssignER"
+        }
+    };
+    std::array<int, num_test_cases> expected_lengths = {
+        {
+            0,
+            4, 4, 8,
+            10, 12, 12
+        }
+    };
 
     for (auto i = 0; i < num_test_cases; i++)
     {
@@ -60,16 +103,7 @@ int main(int, char**)
     }
 
     String s;
-
-    String sFromChar = "word";
-    sFromChar.Print();
-
-    String copyConstructedString = sFromChar;
-    copyConstructedString.Print();
-    sFromChar.Print();
-
-    String moveConstructedString = std::move(sFromChar);
-    moveConstructedString.Print();
+    String moveConstructedString;
 
     String assignmentTester = "yolo";
     assignmentTester.Print();
--- a/course5/homework06/main.cc
+++ b/course5/homework06/main.cc
@@ -58,6 +58,14 @@ int main(int, char**)
     char thechars[] = "chars";
     std::cout << thechars << std::endl;
 
+    std::cout << std::strlen(thechars) << std::endl;
+    const char * res = "";
+    for (auto i = 0; i < 5; i++)
+    {
+        res += thechars[i];
+    }
+    std::cout << res << "  RES" << std::endl;
+
     // char thechars2[] = *p;
     // std::cout << thechars2 << std::endl;
     std::unique_ptr<int[]> int_p = std::make_unique<int[]>(5);
@@ -114,6 +122,13 @@ int main(int, char**)
     }
     assignmentTester.Print();
 
-    assignmentTester[5] = 'c'; // should raise runtime error
+    // assignmentTester[5] = 'c'; // should raise runtime error
+
+    s = "bonanza";
+    std::cout << std::strlen("bonanza") << std::endl;
+    std::cout << s.Length() << std::endl;
+    std::cout << s << std::endl;
+    // std::cout << s.CStr() << std::endl;
+    // assert(s.CStr() == "bonanza");
 
 }
--- a/course5/homework06/main.cc
+++ b/course5/homework06/main.cc
@@ -102,4 +102,18 @@ int main(int, char**)
 
     // How could I use assert to test this one?
     // assignmentTester[5]; // should raise runtime error
+
+    for (auto i = 0; i < 4; i++)
+    {
+        assignmentTester[i] = "word"[i];
+    }
+    assignmentTester.Print();
+    for (auto i = 0; i < 4; i++)
+    {
+        assert(assignmentTester[i] == "word"[i]);
+    }
+    assignmentTester.Print();
+
+    assignmentTester[5] = 'c'; // should raise runtime error
+
 }
--- a/course5/homework06/main.cc
+++ b/course5/homework06/main.cc
@@ -28,7 +28,7 @@ int main(int, char**)
 {
     using namespace simple_string;
     String s;
-    assert(s.Length() == 0);
+    assert(s.length() == 0);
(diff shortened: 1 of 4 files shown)


course5/homework06/main.cc
```diff
@@ -0,0 +1,113 @@
...
+    assignmentTester.Print();
+
+    for (auto i = 0; i < 4; i++)
+    {
+        assert(assignmentTester[i] == "yolo"[i]);
+    }
+    assignmentTester.Print();
+
+    // How could I use assert to test this one?
+    // assignmentTester[5]; // should raise runtime error
```
sophiadavis: I briefly googled for a good way to automate this check. Aside from a test framework, perhaps something clever with try/catch blocks would work? I will check out some collegues' prs...

course5/homework06/main.cc
```diff
@@ -0,0 +1,113 @@
...
+    const char *cStr = s.CStr();
+    std::cout << cStr << std::endl;
+
+    int len = std::strlen(bonanza);
+    for (auto i = 0; i <= len; i++)
+    {
+        assert(cStr[i] == bonanza[i]);
+    }
+
+    // assert(cStr == bonanza); // why does this fail?
```
sophiadavis: I was hoping to replace all my `Print()`'s with `assert(myString.cStr() == "some const char *"`) once I finally figured out cStr (which took me stupid long. I spent literally hours futsing with all manner of looping over strings and `const char []` and forgetting to null-terminate things... But I learned stuff :woman_facepalming:.).

But I never actually got the return value of `CStr()` to `==` an actual `const char *`... What am I doing wrong?

course5/homework06/simple_string.cc
```diff
@@ -0,0 +1,188 @@
...
+
+String::String(String&& src) // move
+{
+    buffer = std::move(src.buffer);
+    size = src.size;
+
+    src.buffer = nullptr;
+    src.size = 0;
+}
+
```
sophiadavis: I figured I didn't need to define a destructor because all member objects (unique_ptr and int) can "clean themselves up".

course5/homework06/simple_string.cc
```diff
@@ -0,0 +1,188 @@
...
+        std::cout << "\"\"" << std::endl;
+    }
+}
+
+int String::Length() const
+{
+    return size;
+}
+
+const char *String::CStr() const
```
sophiadavis: Maybe this is coming from a python place of "there is only one way to do it"-ness, but I'm still really confused about naming conventions (when to camel-case vs snake case vs ... ?).

course5/homework06/simple_string.h
```diff
@@ -0,0 +1,38 @@
...
+    char operator[](const int) const;
+    char& operator[](const int);
+
+    void Print() const;
+    int Length() const;
+    const char *CStr() const;
+
+private:
+    std::unique_ptr<char[]> buffer;
+    int size = 0;
```
sophiadavis: Could have done without the extra bookkeeping of the `size` member, I guess.

course5/homework06/main.cc
maciekgajewski: Simple 'assert' works fine:
```
String a;
assert(s.Length() == 0);

const String b("abc");
assert(b.Length() == 3);
assert(b[0] == 'a');
// etc
```

course5/homework06/main.cc
maciekgajewski: Q: in the following expression, what are the types of rhs and lhs operands? When are they equal?
```
string.c_str() == "text"
```

course5/homework06/simple_string.h
maciekgajewski: Q: What are the pros/cons of doing the extra bookkeeping? Ultimately, which one is better?

course5/homework06/main.cc
```diff
@@ -0,0 +1,113 @@
...
+
+    std::cout << s << std::endl;
+    String emptyStr;
+    std::cout << emptyStr << std::endl;
```
maciekgajewski: Add tests for constant object. Repeat all the tests above that you would expect to work on const string, run them on const string.

course5/homework06/simple_string.h
```diff
@@ -0,0 +1,38 @@
...
+    String(const char*);
+    String(const String& src);
+    String(String&& src);
+    String& operator=(const String& src);
+    String& operator=(String&& src);
+    String& operator=(const char * src);
+    String operator+(const char * src);
+    String operator+(const String& src);
+    String& operator+=(const char * src);
+    String& operator+=(const String& src);
```
maciekgajewski, line 21: Comparison operators are missing (==, !=)

course5/homework06/simple_string.h
```diff
@@ -0,0 +1,38 @@
...
+
+private:
+    std::unique_ptr<char[]> buffer;
+    int size = 0;
+
+}; // String
+
+} // namespace
+
+std::ostream& operator<<(std::ostream&, const simple_string::String&);
```
maciekgajewski: This really should be inside of the namespace. If you struggle with the linking errors, then the underlying problem should be solved.

course5/homework06/simple_string.cc
```diff
@@ -0,0 +1,188 @@
...
+#include <stdexcept>
+
+
+using namespace simple_string;
+
+String::String() = default;
+
+String::String(const char* initString)
+{
+    size = std::strlen(initString);
```
maciekgajewski: include ```<cstring>``` if you're using these

course5/homework06/simple_string.cc
maciekgajewski: And this is a sing that you are using correct tools: not having to implement default constructor and destructor.

course5/homework06/main.cc
```diff
@@ -0,0 +1,113 @@
...
+    moveConstructedString.Print();
+
+    assignmentTester = "yolo";
+    assignmentTester.Print();
+
+    for (auto i = 0; i < 4; i++)
+    {
+        assert(assignmentTester[i] == "yolo"[i]);
+    }
+    assignmentTester.Print();
```
maciekgajewski: Expand the test to cover corner cases:
- Assign empty string to empty string
- Assign non-empty string to empty string
- Assign empty string to non empty string
(repeat for copy/move, and costruction)
Check the content and the length after every operation.
You can implement the test as one function, and array of input value and expected outputs.

course5/homework06/simple_string.cc
```diff
@@ -0,0 +1,188 @@
...
+    else
+        throw std::runtime_error("Index error: '" + std::to_string(i) + "' is out of range.");
+}
+
+char& String::operator[](const int i)
+{
+    if (buffer && i < size)
+        return buffer[i];
+    else
+        throw std::runtime_error("Index error: '" + std::to_string(i) + "' is out of range.");
```
maciekgajewski: ```std::range_error``` would be a better fit.

course5/homework06/simple_string.cc
maciekgajewski: There is no "official" convention.
But in the assignment I specifically asked for source-compatibility with std::string, which means: your method names, param types, return types and behaviour should be the same/compatible with std::string's

maciekgajewski (changes requested):
Mostly good.
There are bugs that are undetected, because the tests doesn't cover all the corner cases.
Tests for cons object missing.
Tests should be automated.

course5/homework06/main.cc
sophiadavis: Hm, interesting, apparently the rhs is a "string literal" -- this causes a compiler warning (`warning: result of comparison against a string literal is unspecified (use strncmp instead)`). 

Thanks

course5/homework06/main.cc
```diff
@@ -0,0 +1,266 @@
...
+    test("concat initialization w String", concatToString, "Hello World Again!");
+
+    concatToString += " And again.";
+    test("+= w const char *", concatToString, "Hello World Again! And again.");
+
+    String concatStr3 = " And again twice.";
+    concatToString += concatStr3;
+    test("+= w String", concatToString, "Hello World Again! And again. And again twice.");
+
+    /* Test out of bounds error with indexing */
```
sophiadavis, line 143: This seems to cause a memory leak -- but unique ptrs are deallocated up in try/catch, I thought?

course5/homework06/main.cc
```diff
@@ -0,0 +1,266 @@
...
+    // std::cout << outcome << test_name << std::endl;
+
+    /* Test assignment & construction corner cases */
+    /*
+    Assign empty string to empty string
+    Assign non-empty string to empty string
+    Assign empty string to non empty string
+    (repeat for copy/move, and construction)
+    Check the content and the length after every operation.
+    You can implement the test as one function, and array of input value and expected outputs.
```
sophiadavis, line 179: I must have missed the corner case you're hinting at -- all the tests I wrote just passed.

course5/homework06/main.cc
maciekgajewski: Yes, but only if there is a ```catch``` block. Otherwise the program will just crash.
But if you will catch your exception, then there should be no leak.

course5/homework06/simple_string.cc
```diff
@@ -0,0 +1,232 @@
...
+}
+
+const char *String::c_str() const
+{
+    const char *toReturn = "";
+    if (buffer)
+    {
+        toReturn = buffer.get();
+    }
+    return toReturn;
```
maciekgajewski: I would write it like this:
```
if (buffer)
    return buffer.get();
else
   return "";
```
It would generate exactly the same code, but seems easier to read.
The roundabout way you choose seems to suggest that you are not entirely sure that ```buffer.get()``` and ```""``` are of the same type.

course5/homework06/simple_string.cc
```diff
@@ -0,0 +1,232 @@
...
+
+namespace simple_string
+{
+
+String::String() = default;
+
+String::String(const char* initString)
+{
+    l = std::strlen(initString);
+    buffer = std::make_unique<char[]>(l + 1);
```
maciekgajewski, line 19: Even if ```l``` is 0?
so:
```
String empty; // doesn't allocate
String empty2 = ""; // allocates
```

course5/homework06/simple_string.cc
```diff
@@ -0,0 +1,232 @@
...
+    l = src.l;
+
+    src.buffer = nullptr;
+    src.l = 0;
+}
+
+String& String::operator=(const String& src) // copy assignment
+{
+    l = src.l;
+    if (src.buffer)
```
maciekgajewski, line 46: So this is the corner case I was looking for.

If you assign empty string to a non-empty one, and the empty string has no buffer, then the assigned-to will keep its old content.

course5/homework06/main.cc
```diff
@@ -0,0 +1,266 @@
...
+    };
+    std::array<String, num_test_cases> inputs = {
+        {
+            defaultS,
+            sFromChar, copyConstructed, moveConstructed,
+            toBeAssigned, copyAssigned, moveAssigned,
+            copyAssigner
+        }
+    };
+    std::array<const char *, num_test_cases> expected_contents = {
```
maciekgajewski, line 113: Instead of having multiple arrays, why not have one array of structures?

course5/homework06/main.cc
```diff
@@ -0,0 +1,266 @@
...
+    String copyAssigner = "CopyAssignER";
+    String moveAssigner = "MoveAssignER";
+    String copyAssigned;
+    copyAssigned = copyAssigner;
+    String moveAssigned;
+    moveAssigned = std::move(moveAssigner);
+
+    copyAssigner[2] = 'z';
+
+    const int num_test_cases = 8;
```
maciekgajewski, line 89: you don't need the constant if you create your array with ```std::make_array```

maciekgajewski (commented):
Nice progress

course5/homework06/main.cc
sophiadavis: I do have a `catch` block. I think it's something about the index-assignment out of bounds?
```
==1508== HEAP SUMMARY:
==1508==     in use at exit: 26,925 bytes in 193 blocks
==1508==   total heap usage: 308 allocs, 115 frees, 33,706 bytes allocated
==1508== 
==1508== 16 bytes in 1 blocks are definitely lost in loss record 4 of 65
==1508==    at 0x1000175B9: calloc (in /usr/local/Cellar/valgrind/3.11.0/lib/valgrind/vgpreload_memcheck-amd64-darwin.so)
==1508==    by 0x1000F8636: __cxa_get_globals (in /usr/lib/libc++abi.dylib)
==1508==    by 0x1000F8A19: __cxa_throw (in /usr/lib/libc++abi.dylib)
==1508==    by 0x1000052DE: simple_string::String::operator[](int) (simple_string.cc:181)
==1508==    by 0x1000069A2: main (main.cc:150)
==1508== 
==1508== 272 bytes in 1 blocks are definitely lost in loss record 50 of 65
==1508==    at 0x100016EBB: malloc (in /usr/local/Cellar/valgrind/3.11.0/lib/valgrind/vgpreload_memcheck-amd64-darwin.so)
==1508==    by 0x10023EB58: getPerThreadBufferFor_dlerror(unsigned long) (in /usr/lib/system/libdyld.dylib)
==1508==    by 0x7FFF5FC0C0C4: dlerrorSet(char const*) (in /usr/lib/dyld)
==1508==    by 0x7FFF5FC0C3F2: dlopen (in /usr/lib/dyld)
==1508==    by 0x10023E79B: dlopen (in /usr/lib/system/libdyld.dylib)
==1508==    by 0x1000FBC40: std::__1::__libcpp_refstring::compute_gcc_empty_string_storage() (in /usr/lib/libc++abi.dylib)
==1508==    by 0x1000FBC0E: std::__1::__libcpp_refstring::get_gcc_empty_string_storage() (in /usr/lib/libc++abi.dylib)
==1508==    by 0x1000FBBB1: std::__1::__libcpp_refstring::~__libcpp_refstring() (in /usr/lib/libc++abi.dylib)
==1508==    by 0x1000FBA51: std::range_error::~range_error() (in /usr/lib/libc++abi.dylib)
==1508==    by 0x1000F8BD2: __cxa_decrement_exception_refcount (in /usr/lib/libc++abi.dylib)
==1508==    by 0x100006A6D: main (main.cc:156)
==1508== 
==1508== 2,064 bytes in 1 blocks are possibly lost in loss record 59 of 65
==1508==    at 0x10001717C: malloc_zone_malloc (in /usr/local/Cellar/valgrind/3.11.0/lib/valgrind/vgpreload_memcheck-amd64-darwin.so)
==1508==    by 0x1005F0EFD: _objc_copyClassNamesForImage (in /usr/lib/libobjc.A.dylib)
==1508==    by 0x1005E4182: protocols() (in /usr/lib/libobjc.A.dylib)
==1508==    by 0x1005E4093: readClass(objc_class*, bool, bool) (in /usr/lib/libobjc.A.dylib)
==1508==    by 0x1005E1C13: gc_init (in /usr/lib/libobjc.A.dylib)
==1508==    by 0x1005E924E: objc_initializeClassPair_internal(objc_class*, char const*, objc_class*, objc_class*) (in /usr/lib/libobjc.A.dylib)
==1508==    by 0x1005F6132: layout_string_create (in /usr/lib/libobjc.A.dylib)
==1508==    by 0x1005E483C: realizeClass(objc_class*) (in /usr/lib/libobjc.A.dylib)
==1508==    by 0x1005E4300: copySwiftV1MangledName(char const*, bool) (in /usr/lib/libobjc.A.dylib)
==1508==    by 0x1005E42E9: copySwiftV1MangledName(char const*, bool) (in /usr/lib/libobjc.A.dylib)
==1508==    by 0x1005E42E9: copySwiftV1MangledName(char const*, bool) (in /usr/lib/libobjc.A.dylib)
==1508==    by 0x1005E42E9: copySwiftV1MangledName(char const*, bool) (in /usr/lib/libobjc.A.dylib)
==1508== 
==1508== LEAK SUMMARY:
==1508==    definitely lost: 288 bytes in 2 blocks
==1508==    indirectly lost: 0 bytes in 0 blocks
==1508==      possibly lost: 2,064 bytes in 1 blocks
==1508==    still reachable: 4,096 bytes in 1 blocks
==1508==         suppressed: 20,477 bytes in 189 blocks
```

course5/homework06/main.cc
sophiadavis: Good to know, thanks!

course5/homework06/main.cc
sophiadavis: Yeah, that would have made it a lot easier... I'll use that approach next time.

course5/homework06/simple_string.cc
sophiadavis: Yeah, isn't that right? "" is still one null character, so length of 1 char.

sophiadavis: I think I'd rather change it so they both still allocate.

course5/homework06/simple_string.cc
sophiadavis: But I allocate a buffer even for "" (now in both default and from-const-char-* constructor), so the assigned-to does get new content.

course5/homework06/simple_string.cc
sophiadavis: Yeah, that's a lot clearer -- thanks.

course5/homework06/main.cc
maciekgajewski: Interesting. It looks like the leaked buffer were allocated somewhere in Darwin's exception throwing mechanism.
To make sure, you should try running the same test on Linux.

PR #134, opened 2017-03-13
homework week 6

anilsukul:


course5/homework6/mystr.h
```diff
@@ -0,0 +1,46 @@
+#pragma once 
+
+#include <string>
+#include <cstdio>
```
maciekgajewski, line 4: I don't see the need for including string or cstdio here

course5/homework6/mystr.h
```diff
@@ -0,0 +1,46 @@
...
+	MyString& operator = (const char* str); // assignment using char*
+
+	MyString(const MyString&& str);  // move constructor
+	MyString& operator = (const MyString&& str);  // move assignment
+	
+	char& operator[] (const size_t index); 
+	const char& operator[] (const size_t index) const; 
+
+	MyString  operator+ (const MyString& str); // concatenation
+	MyString&  operator+= (const MyString& str); // concatenation assignment
```
maciekgajewski, line 34: Please, add operator == and !=. It is essential for testing

course5/homework6/mystr.cc
```diff
@@ -0,0 +1,104 @@
+#include "mystr.h"
+#include <memory>
+#include <cmath> 
+#include <cstring>
+
+using namespace mystring;
+
+MyString::MyString()
+{
+	mStr = std::make_unique<char[]>(1); //length 1 for zero terminator
```
maciekgajewski, line 10: This produces correct result, but it's not ideal. Ideally, the default constructor should not allocate (or do other expensive stuff).
Quite often you will find yourself creating default-constructed object, ad then, depending on runtime conditions, immediately assigning to it, or discarding it.

Consider this:
```
std::string text;
if (dataSource == DataSource::FILE)
   text = ReadFromFile(text);
else
   text = ReadFromMemory();
AnalyseText(text);
```
In the example above you see that the "content" of the default-constructed string is short lived.

course5/homework6/mystr.cc
```diff
@@ -0,0 +1,104 @@
...
+	mStr[0] = '\0';
+}
+
+// Function to create a new unique pointer, memset it and initialize it with a 
+// new string. 
+// Repeatedly needed for copy/move assignment and construction
+void MyString::Reset_mStr(int len, const char* str)
+{
+	mStr = std::make_unique<char[]>(len+1); // add 1 for \0 termination
+	std::memset(mStr.get(), 0, len + 1);
```
maciekgajewski, line 20: Why zero all the bytes, if you are immediately override them with another content?

course5/homework6/mystr.cc
```diff
@@ -0,0 +1,104 @@
...
+}
+
+MyString& MyString::operator = (const MyString& str)  // copy assignment
+{
+	Reset_mStr(str.length(), str.c_str());
+	return *this;
+}
+	
+
+MyString:: MyString(const MyString&& str)  // move constructor
```
maciekgajewski, line 36: This is not move constructor. Move constructor takes non-const rvalue referenvce (```MyString&&```), so that you can modify the source object.
You need to be able to modify the source object to steal from it.
You implementation simply copies from the source, leaving the source intact; it is effectively a copy constructor.

The idea behind move constructor is to avoid O(N) copy, and replace it with O(1) stealing.

course5/homework6/mystr.cc
```diff
@@ -0,0 +1,104 @@
...
+{
+	Reset_mStr(std::strlen(str), str);
+	return *this;
+}
+
+	
+char& MyString::operator[] (const size_t index) 
+{
+	if (index >= this->length())
+		throw std::runtime_error ("Index out of bound");
```
maciekgajewski, line 62: ```std::range_error``` seems like a better fit.

maciekgajewski: Why ```>=```? Is asking for the first character in empty string OK?

course5/homework6/homework06_stdstr.cc
```diff
@@ -0,0 +1,104 @@
...
+#include <iostream>
+#include <cassert>
+#include <string>
+#include <cstring>
+
+void test_operators();
+void test_basics();
+
+
+void test_basics()
```
maciekgajewski, line 11: This test_basic, and the test_basic in the other file are essentially the same. This call for a template! We'll discuss them today.

course5/homework6/homework06_mystr.cc
```diff
@@ -0,0 +1,105 @@
...
+
+void test_operators();
+void test_basics();
+
+using namespace mystring;
+
+void test_basics()
+{
+	std::cout << "====== Testing basics ======" << std::endl;
+	MyString empty("");
```
maciekgajewski, line 15: You never check the properties of default-constructed object

course5/homework6/homework06_mystr.cc
```diff
@@ -0,0 +1,105 @@
...
+
+	MyString str3;
+	str3 = "Haleluja"; // copy assignment with char*
+
+	std::cout << "str3 created and assigned by copy. SHould be Haleluja: "<<  str3 << std::endl; 
+	assert(str3.length() == std::strlen("Haleluja"));
+
+	str3 = str2; // copy assignment with MyString
+	std::cout << "str3. should be back to str2 (Hello World): "<<  str3 << std::endl; 
+	assert(str3.length() == str2.length());
```
maciekgajewski, line 37: Implementing operator == would allow you to test the content of the string.

course5/homework6/homework06_mystr.cc
```diff
@@ -0,0 +1,105 @@
...
+}
+
+void test_operators()
+{
+	char test_str[] = "The Quick Brown Fox Jumps Over The Lazy Dog"; 
+	std::cout << "====== Testing operators ======" << std::endl;
+	MyString str(test_str);
+	MyString str2 = std::move(str);
+
+	std::cout << "Moved String " << str2 <<  ". Len = "<< str2.length() << std::endl;
```
maciekgajewski, line 50: What's the length of str after the move?

course5/homework6/homework06_mystr.cc
```diff
@@ -0,0 +1,105 @@
...
+
+	std::cout << "creating full string with different operator" << std::endl;
+	MyString str("Hello world");
+	assert(str.length() == 11); 
+
+	std::cout << "Created empty string str='"<< empty << "' and full string str='" << str  << "'" << std::endl;
+
+	MyString str2 = str; // copy constructor
+	std::cout << "Created str2 from str= "<<  str2 << std::endl; 
+	assert(str.length() == str2.length());
```
maciekgajewski, line 27: is ```str``` a deep copy of ```str2```? Can you test it?

course5/homework6/homework06_mystr.cc
```diff
@@ -0,0 +1,105 @@
...
+	str5 += str6;
+	std::cout << str5 << std::endl; 
+	assert (str5.length() == 12);
+
+	std::cout << "Testing COntatenation. " ; 
+	MyString str7 = "Hello";
+	MyString str8 = " World Again !";
+	std::cout << str7 + str8 << std::endl;
+	std::cout << "~~~~~ Finished Testing operators" << std::endl;
+}
```
maciekgajewski, line 95: Cover corner cases:
- assigning empty string to non-empty string
- assigning empty string to empty string
- assigning non-empty string to empty string
etc

maciekgajewski (changes requested):
I like the idea of having _similar_test for your string and std::string. Let's use template and make it really _the same_ test.
Move construction/assignment doesn't really move (Should be O(1)).
Add operator ==, !=, expand tests to test the content and cover corner cases.

PR #135, opened 2017-03-13
Hw6

starostinak:


course5/homework6/string.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include <cstring>
+#include <iostream>
+
+class String
+{
+public:
+    String() { }
+
+    String(const String& other)
+        : buffer_(std::make_unique<char[]>(sizeof(other.buffer_)))
```
maciekgajewski, line 11: what will the ```sizeof``` return?

course5/homework6/string.cpp
```diff
@@ -0,0 +1,117 @@
...
+    String d = b + c;
+    std::cout << "b + c: " << d << std::endl;
+    d += "ddd";
+    std::cout << "d: " << d << std::endl;
+
+    const String& dd = d;
+    std::cout << dd[0] << dd[1] << dd[2] << std::endl;
+
+    return 0;
+}
```
maciekgajewski, line 117: The tests are painfully incomplete.
- test const object
- test for strings longer than 8 characters (and see it crash!)
- test for corner cases: assignment/copy from empty, to empty, adding empty strings etc

course5/homework6/string.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    String(const String& other)
+        : buffer_(std::make_unique<char[]>(sizeof(other.buffer_)))
+        , size_(other.size_)
+    {
+        std::uninitialized_copy_n(other.buffer_.get(), size_, buffer_.get());
+    }
+
+    String(const char* str)
+        : buffer_(std::make_unique<char[]>(sizeof(str)))
```
maciekgajewski, line 18: :cold_sweat:

course5/homework6/string.cpp
```diff
@@ -0,0 +1,117 @@
...
+        , size_(std::strlen(str))
+    {
+        std::uninitialized_copy_n(str, size_, buffer_.get());
```
maciekgajewski, line 21: uninitialized_copy makes little sense here. std::memcpy would suffice

course5/homework6/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
+set(CMAKE_CXX_STANDARD 14)
+
+add_executable(hw6 string.cpp)
```
maciekgajewski, line 4: Split the code into at least 2 files

course5/homework6/string.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    String(String&& other)
+        : buffer_(std::move(other.buffer_))
+        , size_(other.size_)
+    {
+        other.size_ = 0;
+    }
+
+    String& operator=(const String& other) = default;
+    String& operator=(const char* str) // I have no idea why =default doesn't compile
```
maciekgajewski, line 32: because assignment from different type is not one of the Big 6

course5/homework6/string.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    String(String&& other)
+        : buffer_(std::move(other.buffer_))
+        , size_(other.size_)
+    {
+        other.size_ = 0;
+    }
+
+    String& operator=(const String& other) = default;
```
maciekgajewski, line 31: Does it compile? How does it work?

course5/homework6/string.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+private:
+    std::unique_ptr<char[]> buffer_;
+    size_t size_ = 0;
+};
+
+std::ostream& operator<<(std::ostream& os, const String& str)
+{
+    if (str.length()) {
```
maciekgajewski, line 86: Why is this needed?

maciekgajewski (changes requested):
Many things is wrong with this code

course5/homework6/string.cpp
maciekgajewski: Ha! I knew it! I doesn't actually compile. The caveat is: the compiler will try to generate the function only when you are using it. And you are not, which exposes a hole in your test coverage.

PR #137, opened 2017-03-14
Homework6

jfdsmit:


course5/homework6/string.h
```diff
@@ -0,0 +1,34 @@
+#ifndef _JDS_STRING_H_
```
maciekgajewski: ```#pragma once``` is better

course5/homework6/string.h
```diff
@@ -0,0 +1,34 @@
...
+	void operator=(String const & that);
+	void operator=(String && that);
+	void operator=(char const * source);
+	unsigned length() const;
+	char& operator[](unsigned index);
+	char operator[](unsigned index) const;
+	String operator+(String const & that) const;
+	String& operator+=(String const & that);
+	char const * c_str() const;
+
```
maciekgajewski: I've realised that operators == and != are very useful in automating the unit test

course5/homework6/string.h
```diff
@@ -0,0 +1,34 @@
...
+private:
+	std::unique_ptr<char[]> mBuffer;
+	unsigned mLength;
+};
+
+}
+
+std::ostream& operator<<(std::ostream& s, jds::String const & value);
```
maciekgajewski: There is no reason for it to be in global namespace. You should be able to leave in the class' namespace.

course5/homework6/string.cc
```diff
@@ -0,0 +1,115 @@
+#include <memory>
+#include <cstring>
+
+#include "string.h"
+
+namespace jds {
+
+String::String()
+{
+	mLength = 0;
```
maciekgajewski: Just set the default value in the class definition, and get away with compiler-generated constructor.

course5/homework6/string.cc
```diff
@@ -0,0 +1,115 @@
...
+{
+	mLength = that.mLength;
+	mBuffer = std::make_unique<char[]>(mLength + 1);
+	std::memcpy(this->mBuffer.get(), that.mBuffer.get(), that.mLength);
+	mBuffer.get()[mLength] = '\0';
+}
+
+String::String(String && that)
+{
+	mBuffer = std::unique_ptr<char []>(that.mBuffer.release());
```
maciekgajewski: Ugly! :scream_cat:
Use ```std::move``` instead

course5/homework6/string.cc
```diff
@@ -0,0 +1,115 @@
...
+	mLength = that.mLength;
+	that.mLength = 0;
+}
+
+String::String(char const * source)
+{
+	mLength = strlen(source);
```
maciekgajewski: ```std::strlen```

course5/homework6/string.cc
```diff
@@ -0,0 +1,115 @@
...
+
+String::String()
+{
+	mLength = 0;
+}
+
+String::String(String const & that)
+{
+	mLength = that.mLength;
+	mBuffer = std::make_unique<char[]>(mLength + 1);
```
maciekgajewski: So, copying from an empty string will allocate buffer of 1?

course5/homework6/string.cc
```diff
@@ -0,0 +1,115 @@
...
+{
+	mBuffer = std::unique_ptr<char []>(that.mBuffer.release());
+	mLength = that.mLength;
+	that.mLength = 0;
+}
+
+void String::operator=(char const * source)
+{
+	mLength = strlen(source);
+	mBuffer = std::make_unique<char[]>(mLength + 1);
```
maciekgajewski: Even doing this:
```
s = "";
```
will allocate.

course5/homework6/string.cc
```diff
@@ -0,0 +1,115 @@
...
+}
+
+unsigned String::length() const
+{
+	return mLength;
+}
+
+char& String::operator[](unsigned index)
+{
+	if (index >= mLength)
```
maciekgajewski, line 69: why ```>=```?
What if your string is default-constructed?

course5/homework6/string.cc
```diff
@@ -0,0 +1,115 @@
...
+	{
+		throw std::runtime_error("String index out of bounds");
```
maciekgajewski: ```std::range_error``` may be better.

course5/homework6/main.cc
```diff
@@ -0,0 +1,99 @@
...
+	assert(std::strcmp((assigned_cstr + test_string).c_str(), "abcdef") == 0);
+	pass();
+
+	announce("operator+=");
+	assigned_cstr += test_string;
+	assert(std::strcmp(assigned_cstr.c_str(), "abcdef") == 0);
+	pass();
+
+	return 0;
+}
```
maciekgajewski: - tests fro const object are missing
- test that copying is indeed deep (modyfying copy doesn't modify the source)
- test corner cases - assigning from empty, to empty, adding two empty etc

maciekgajewski (changes requested):
Nice and clean code, but:
- unnecessary allocations
- expand tests

course5/homework6/string.cc
jfdsmit: Yes it would and no it should not do that

course5/homework6/string.cc
jfdsmit: A string with length x has only valid indices 0 through x-1, hence if index == length it's already out of range. If the string is default constructed it's empty and has no accessible characters.

If you don't like >= and <= the logic could be inverted to operate normally when index < mLength and error out in the other branch.

maciekgajewski: yes, of course, it is correct. My bad

course5/homework6/string.h
jfdsmit: Yes, I stuck too religiously to the listing in the e-mail

course5/homework6/string.cc
jfdsmit: Tried that first last night and did not get it to compile for some reason so I thought I was doing something wrong. This worked too. Your assertion that `std::move` should work made me put it back in and apparently I did something wrong yesterday because it indeed "just works".

course5/homework6/main.cc
jfdsmit: I'm not exactly sure what "tests for const object" implies, could you elaborate what should be tested about const objects?

maciekgajewski:
```
const String cabc = "abc";
assert(cabc.length() == 3);
assert(cabc[1] = 'b');
assert(cabc == "abc");
// etc..
```

PR #141, opened 2017-04-02
Homework 9-2

sophiadavis:
It works with no memory leaks for the given tests.

I wasn't kind of surprised to find that move construction/assignment from instances w type of `const char *` to instances w type `std::string` didn't compile.

### DIFF
--- a/course5/homework09_2/main.cc
+++ b/course5/homework09_2/main.cc
@@ -31,7 +31,6 @@ struct MoreUniquePtr
 			delete ptr;
 	}
 
-	// copy assignment operator
 	MoreUniquePtr& operator=(const MoreUniquePtr& src) = delete; // copy assignmt
 
 	MoreUniquePtr& operator=(MoreUniquePtr&& src) // move assignmt
--- a/course5/homework09_2/main.cc
+++ b/course5/homework09_2/main.cc
@@ -1,0 +1,82 @@
+/*
+Write a simple single-object implementation of std::unique_ptr, complete with make_unique.
+*/
+#include <iostream>
+#include <string>
+#include <cassert>
+
+
+// operators new/delete or new[]/delete[]
+
+template<typename T>
+struct MoreUniquePtr
+{
+	MoreUniquePtr() = default;
+	MoreUniquePtr(const T& src)
+	{
+		ptr = new T;
+		*ptr = src;
+	}
+	MoreUniquePtr(const MoreUniquePtr&) = delete; // copy
+
+	// MoreUniquePtr(MoreUniquePtr&& src) // move
+	// {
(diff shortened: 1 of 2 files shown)


course5/homework09_2/main.cc
```diff
@@ -0,0 +1,116 @@
...
+*/
+#include <iostream>
+#include <string>
+#include <cassert>
+
+
+// operators new/delete or new[]/delete[]
+
+template<typename T>
+struct MoreUniquePtr
```
maciekgajewski, line 12: Oh c'mon! You could've put this into it's own header file.

maciekgajewski: And whats with the name?

course5/homework09_2/main.cc
```diff
@@ -0,0 +1,116 @@
...
+{
+	MoreUniquePtr() = default;
+	MoreUniquePtr(const T& src)
```
maciekgajewski, line 15: Is this what std::unique_ptr is doing?

And, such a constructor, not marked as 'explicit', would be very dangerous, could lead to unintended code to compile.

course5/homework09_2/main.cc
```diff
@@ -0,0 +1,116 @@
...
+	{
+		if (ptr)
+			delete ptr;
+	}
+
+	MoreUniquePtr& operator=(const MoreUniquePtr& src) = delete; // copy assignmt
+
+	MoreUniquePtr& operator=(MoreUniquePtr&& src) // move assignmt
+	{
+		if (ptr)
```
maciekgajewski, line 38: It's safe to call ```delete``` on a null pointer. The branch is redundant.

course5/homework09_2/main.cc
```diff
@@ -0,0 +1,116 @@
...
+			delete ptr;
+		ptr = src.ptr;
+		src.ptr = nullptr;
+		return *this;
+	}
+
+	T get()
```
maciekgajewski, line 45: This is not what std:unique_ptr is doing.

The semantic you are proposing is the one of a single-item container rather than smart pointer.
You provide functionality to copy data in or out of it. It is almost tempting to add a copy constructor that copies the content between the instances (just like containers do).

Smart pointer simply provides basic memory management while still retaining pointer semantics. It is still a "handle to an object", not a container. User has direct access to "the object" via ```get()```, ```operator*``` and ```operator->```.

maciekgajewski (changes requested):
This is not a smart pointer!

- put your code in a separate header; try to keep your code to a standard you'd expect from a production ready library.
- study the interface of std::unique_ptr
- implement (variadic!) make_unique

PR #142, opened 2017-04-03
Dani Homework 6

vargad:


course5/homework06/homework06.cc
```diff
@@ -0,0 +1,596 @@
...
+auto testStrings = {
+    std::make_tuple("", 0),
+    std::make_tuple("A", 1),
+    std::make_tuple("AB", 2),
+    std::make_tuple("ABC", 3),
+    std::make_tuple("abcd", 4),
+    std::make_tuple("Mokus", 5),
+    std::make_tuple("Macska", 6),
+    std::make_tuple("Macs\0ka", 7),
+    std::make_tuple(u8"A kis mókus felmászott a fára és sok mogyorót talált, amit elrejtett, hogy legyen mit ennie a télen.", 107)
```
maciekgajewski, line 536: You count the accented charters double?

course5/homework06/homework06.cc
```diff
@@ -0,0 +1,596 @@
...
+    compare_test<operator_plus_equal_char>();
+    compare_test<operator_plus_equal_cstring>();
+    compare_test<operator_plus_equal_string>();
+    compare_test<operator_plus_char>();
+    compare_test<operator_plus_char_nul>();
+    compare_test<operator_plus_cstring>();
+    compare_test<operator_plus_string>();
+    compare_test<operator_plus_string_with_nul>();
+    compare_test<push_back>();
+    compare_test<to_stream>();
```
maciekgajewski, line 595: Fancy.

Too fancy. Fun to write, but not fun to read, and rather rigid.
You could've achieve the same result with simple functions templates, without tuples and functors.

course5/homework06/ustring.h
```diff
@@ -0,0 +1,121 @@
...
+    String operator+(const char *str) const;
+    String operator+(const String &str) const;
+    String& operator+=(char c) { push_back(c); return *this; }
+    String& operator+=(const char *str);
+    String& operator+=(const String &str);
+    void resize(std::size_t size);
+    void resize(std::size_t size, char c);
+
+private:
+    static char emptyString[];
```
maciekgajewski, line 105: This is not really needed.
1. It should be ```const char*```, not ```char*```
2. no reason to prefer the more obscure ```char v[]``` notation, over ```char* v```
3. Code would be much simpler if you'd simply use ```""```` when appropriate.

course5/homework06/ustring.h
```diff
@@ -0,0 +1,121 @@
...
+    char& front() { return (*this)[0]; }
+    const char& front() const { return (*this)[0];  }
+
+    char& back() { return (*this)[mLength-1]; }
+    const char& back() const { return (*this)[mLength-1]; }
+
+    char* data()
+    {
+        if (length() == 0)
+            return emptyString;
```
maciekgajewski, line 55: This is ugly , and a lie. You offer a writable access to a buffer that doesn't really exist.
It would be much cleaner to simply return null pointer for empty string.

course5/homework06/ustring.h
```diff
@@ -0,0 +1,121 @@
...
+class String
+{
+public:
+    static constexpr std::size_t npos = -1;
+
+    String() = default;
+
+    String(const char *str) { *this = str; }
+    String(const char *str, size_t length);
+    String(String const &str) { *this = str; }
```
maciekgajewski, line 20: I find this syntax for const references less readable, as it is inconsistent with const pointer syntax.

maciekgajewski (commented):
Good.

There was really no need to implement the reusable buffer, but you did it and seems to work.
The test is waaay to complex. It could be 33% of the code.

course5/homework06/ustring.h
vargad: Since C++11 std::string::data() must return a pointer to single null character when length is zero: http://en.cppreference.com/w/cpp/string/basic_string/data
Yes, it's a lie, because it's write access to something, but there is really no place to write to. I wonder how the standard library implements this.

maciekgajewski: ```std::string::data() const```, yes. It's essentially equivalent to c_str().
The non-const version will be added only in C++17

course5/homework06/homework06.cc
vargad: It's UTF8, I counted the actual number of bytes, in this case accented characters take 2 bytes, indeed.

course5/homework06/ustring.h
vargad: It's not the same thing. What I did is a global (writable) buffer with only a zero byte. With const char* it's a pointer to a read only string literal with only a zero byte.
https://godbolt.org/g/3nRS8p
https://godbolt.org/g/Z921Sq
I need "writable" buffer since data() must be able to return an empty buffer, or I need to dynamically allocate space for empty string as well.

course5/homework06/ustring.h
vargad: Well, that explains why I had to remove a check when I changed the code to work on C++14. I was wondering why the data() returns the same pointer after coping an std::string, but too sleepy to look at it closely. I pre gcc 5 std::string is copy on write, and data() is const so copy constructor doesn't actually have to copy anything at all.

course5/homework06/ustring.h
maciekgajewski: You'd need writable buffer only if you want to implement non-const data.
IMO the best solution is to remove the non-const data().

If you insist on having non-const data, that I don't think returning pointer to global buffer is a good idea; I think other solution would be better; I don't know what C++17 string is doing, thou...

PR #143, opened 2017-04-03
Dani Homework 6 with small string optimization

vargad:


course5/homework06/ustring.h
```diff
@@ -0,0 +1,138 @@
...
+        SmallString(const char *s, std::size_t length)
+        {
+            assert(length < SmallStringSize);
+            std::copy(s, s+length, str);
+        }
+        char str[SmallStringSize];
+    };
+    std::size_t mCapacity = SmallStringSize-1;
+    std::size_t mLength = 0;
+    std::variant<SmallString, std::unique_ptr<char[]>> mStr = SmallString();
```
maciekgajewski, line 128: nah! In case of a small string you should be able to reuse at least the space used by ```mCapacity```, because you don't need it in suxch a case.
Depending on your semantic, you may also be able to write over ```mLength```.

I'm also not sure if ```std::variant``` is a good choice; it's a tagged union, which means it has extra field to store the content type.

Please note: gcc5's std::string implements SSO for strings <24 characters, and the total size of the object is 32.
What is the size of your string now?

maciekgajewski (commented):
Not very efficient, and the size of the string object is now quite enormous, i suspect.

PR #145, opened 2017-04-05
Fei homework6

feicui08:


course5/homework06/string.cc
```diff
@@ -0,0 +1,150 @@
+#include "string.h"
+#include<memory>
+#include<algorithm>
+#include<cstring>
+
+String::String()
```
maciekgajewski, line 6: This constructor leaves ```size``` uninitialised.
You should either provide the default value in the class' definition (better option), or set the default value here (less nice option)

course5/homework06/string.cc
```diff
@@ -0,0 +1,150 @@
...
+
+String::String(const String& src)   //copy constructor
+{
+    size = src.length();
+    charbuffer = std::make_unique<char[]>(size+1);
+    //for (int i=0; i < size+1; ++i)
+    //{
+        //charbuffer[i] = src.charbuffer[i];
+    //}
+    //std::copy(&(src.charbuffer[0]), &(src.charbuffer[size+1]), &(charbuffer[0])); //because charbuffer is unique pointer, otherwise if it's a char*, I can just use charbuffer
```
maciekgajewski, line 19: It's nice to see you experimenting with different solution, and I'm happy that you settled for the best one :)
But we have version control (git), and you can safely remove the commented-out code.

course5/homework06/string.cc
```diff
@@ -0,0 +1,150 @@
...
+String::String(const char* src)     //converting constructor
+{
+    *this = src;
+}
+
+String& String::operator= (const char* src)     //converting assignment
+{
+    size = std::strlen(src);
+    charbuffer = std::make_unique<char[]>(size+1);
+    for (int i=0; i < size+1; ++i)
```
maciekgajewski, line 55: Why not use std::copy here?

course5/homework06/string.h
```diff
@@ -0,0 +1,39 @@
+#include<memory>
```
maciekgajewski, line 1: Start every header file with ```#pragma once```! always!

course5/homework06/string.h
```diff
@@ -0,0 +1,39 @@
...
+#include<iostream>
+
+class String
```
maciekgajewski, line 4: Enclose your code in a namespace

course5/homework06/homework06.cc
```diff
@@ -0,0 +1,57 @@
+#include<iostream>
+#include <cassert>
+#include<cstring>
+#include "string.h"
+
+int main()
+{
+    String mystr1 = "hello";
+    std::string str1 = "hello";
```
maciekgajewski, line 9: Instead of doubling every test, write your test as a template function, and run it once for std::String, and once for your string.

maciekgajewski (commented):
Good!

The logic is good. I have few comments regarding the style.

PR #149, opened 2017-04-09
Hw10 linked

meszenab:


course5/homework10/llist.h
```diff
@@ -0,0 +1,175 @@
...
+
+namespace course 
+{
+template<class T>
+struct node
+{
+//public:
+        node() { /*next.get() = nullptr;*/ }
+        
+ /*       class iterator
```
maciekgajewski: Stop leaving the commented-out code around!

course5/homework10/llist.h
```diff
@@ -0,0 +1,175 @@
+#pragma once
+
+#include <memory>
+#include <utility>
+
+namespace course 
+{
+template<class T>
+struct node
```
maciekgajewski: So ```node``` is a class on equal terms with ```llist```, which makes it a part of the public API. This is probably not what you want.

I suggest you make node a nested class inside of llist.

course5/homework10/llist.h
```diff
@@ -0,0 +1,175 @@
...
+                }
+                else
+                {
+                    throw std::range_error("End of list");
+                }
+            }
+            T& operator*() { return node_ptr->value; }
+            node<T>* node_ptr;
+        };
+        llist() : len(0) { first = std::make_unique<node<T>>(); }
```
maciekgajewski: Why do you allocate the node, if there's nothing to store in it?
What if T is not default-constructible?
Extra task: test your list with not-default-constructible T

course5/homework10/llist.h
```diff
@@ -0,0 +1,175 @@
...
+            for( auto i = other.begin(); i != other.end(); ++i )
+            {
+                new_node->value = *i;
+                new_node->next = std::make_unique<node<T>>();
+                new_node = new_node->next.get();
+            }
+            return *this;
+        }
+
+        llist(const iterator& it1, const iterator& it2)
```
maciekgajewski: This only allows for intializing list from a range on another list of the same type.
I'd like to be able to initialize the list from any iterator range, even from another container type.

course5/homework10/llist.h
```diff
@@ -0,0 +1,175 @@
...
+        {
+            first = std::make_unique<node<T>>();
+            len = 0;
+            node<T>* new_node = first.get();
+
+            for( auto i = it1; i != it2; ++i )
```
maciekgajewski: This function, and copy constructor, and operator=, all share the same code. Why don't you put it into a separate function that you can re-use?
Less code = less bugs.
I suggest ```llist::assign (It first, It last)```, just like in ```std::list``` and other containers.

course5/homework10/llist.h
```diff
@@ -0,0 +1,175 @@
...
+                new_node.get()->next = std::move(first);
+                first = std::move(new_node);
+            }
+            first.get()->value = val;
+            len++;
+        }
+
+        void pop_front()
+        {
+            if (len)
```
maciekgajewski: Q: Is this ```if```` really required? What are the pros/cons of having it?

course5/homework10/llist.h
```diff
@@ -0,0 +1,175 @@
...
+                len--;
+            } else 
+            {   
+                throw std::range_error("List is empty");
+            }
+        }
+
+        T& front() 
+        { 
+            if(len)
```
maciekgajewski, line 142: Same Q as above applies to this ```if```

course5/homework10/llist.h
```diff
@@ -0,0 +1,175 @@
...
+             return it;
+        }
+        iterator end()
+        { 
+             iterator it;
+             return it;
+        }
+        const iterator end() const 
+        {
+            iterator it;
```
maciekgajewski: what's the point in creating all these local variables, here and other bige/end implementations?

course5/homework10/llist.h
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        iterator begin() 
+        {
+             iterator it( first.get() );
+             return it;
+        }
+        const iterator begin() const
+        {
+             iterator it( first.get() );
```
maciekgajewski: This is a const method, and yet  I could be able to modify the content of the container via the iterator it returns.
This breaks the const-correctness.

course5/homework10/llist.h
```diff
@@ -0,0 +1,175 @@
...
+             iterator it;
+             return it;
+        }
+        const iterator end() const 
+        {
+            iterator it;
+            return it;
+        }
+
+        std::unique_ptr<node<T>> get_first() { return first; }
```
maciekgajewski: Why would you provide such a method in a public interface? What is the use-case?

course5/homework10/main.cc
```diff
@@ -0,0 +1,58 @@
...
+    auto list5 = list3;
+    std::for_each(list3.begin(), list3.end(), square);
+    assert( list3.front() == list5.front() * list5.front() );
+    list3.pop_front();
+    list5.pop_front();
+    assert( list3.front() == list5.front() * list5.front() );
+    list3.pop_front();
+    list5.pop_front();
+    assert( list3.front() == list5.front() * list5.front() );
+}
```
maciekgajewski: Tests on const objects are missing. Some of the functionality should work on const objects as well.

course5/homework10/llist.h
```diff
@@ -0,0 +1,175 @@
...
+                {
+                    len++;
+                    new_node->value = *i;
+                    new_node->next = std::make_unique<node<T>>();
+                    new_node = new_node->next.get();
+                }
+            }
+        }
+
+        void push_front(const T& val)
```
maciekgajewski, line 110: There should also be moving push_front, and emplace_front

course5/homework10/main.cc
maciekgajewski: Test if your container works with range-based for loop

maciekgajewski: Test the list with something more complex than int. Test it with non-copyable type, and also with non-default-constructible object.

maciekgajewski (changes requested):
Quite good, but incomplete:

- tests are too simple; test corner cases, test the list with more complex types with exotic properties.
- const correctness is broken
- some methods are missing (push_front(T&&&), emplace_front)
- the code is too complex. You have 3 methods that assign to the list, all doing the same, non-trivial thing
- you keep leaving commented-out code around. This is not acceptable in prod-quality codebase

course5/homework10/llist.h
```diff
@@ -0,0 +1,184 @@
...
+        struct node
+        {
+                node() { }
+                T value;
+                std::unique_ptr<node> next;
+        };
+        std::unique_ptr<node> first;
+        int len = 0;
+public:
+ /*       struct iterator
```
maciekgajewski: Don't keep the commented-out code around.

course5/homework10/llist.h
```diff
@@ -0,0 +1,184 @@
...
+                }
+                else
+                {
+                    throw std::range_error("End of list");
+                }
+            }
+            T& operator*() { return node_ptr->value; }
+            node* node_ptr;
+        };*/
+        template<class P, class Q>
```
maciekgajewski: This class is already a template by the virtual of being nested under one (the fully-qualified name of the class is ```course::llist<T>::iterator_class```. There is really no need to make it even more templaty.

course5/homework10/llist.h
```diff
@@ -0,0 +1,184 @@
...
+            for( auto i = other.begin(); i != other.end(); ++i )
+            {
+                new_node->value = *i;
+                new_node->next = std::make_unique<node>();
+                new_node = new_node->next.get();
+            }
+            return *this;
+        }
+
+        template<typename Q>
```
maciekgajewski: In case like this, it's better to use some descriptive name for the type, rather than Q. Maybe ```Iterator``` or '''It``` ?

maciekgajewski (changes requested):
- Non need to make iterator a template, it is already one.
- emplace_front is still missing
- you still keep commented-out code

course5/homework10/llist.h
meszenab: The reason I have done in this way is to implement both a (non-const) iterator and a const_iterator class. Since the code for the two would be almost identical I thought this is an elegant solution with
iterator = iterator_class<node*, T> and const_iterator = iterator_class<const node*, const T>.
Is there a simpler way to achieve this (implement iterator and const_iterator without code duplication)?

PR #150, opened 2017-04-10
Homework06

AJIEIIIA:


course5/homework06/main.cpp
```diff
@@ -0,0 +1,31 @@
+#include <iostream>
+#include <cassert>
+#include "string.h"
+#include "test.h"
```
maciekgajewski: Wrong order of includes. Always put local headers first.

course5/homework06/string.cc
```diff
@@ -0,0 +1,120 @@
+#include <cstring>
+#include "string.h"
```
maciekgajewski: swap the order

course5/homework06/string.h
```diff
@@ -0,0 +1,43 @@
...
+#include <ostream>
+#include <cstdlib>
+
+
+class String
+{
+public:
+    String() = default;
+    String(String &&);
+    String(char* data);
```
maciekgajewski: Is this constructor needed? Any ```char*``` can be implicitly cast to ```const char*```, and you are not modifying the content of the buffer.

course5/homework06/string.h
```diff
@@ -0,0 +1,43 @@
...
+    String(const char* data);
+    String(const String& data);
+
+    String& operator = (const String& string);
+    String& operator = (String && string);
+    String& operator = (const char* data);
+
+
+    char& operator [](std::size_t index);
+    const char& operator [] (std::size_t index) const;
```
maciekgajewski, line 21: What's cheaper: returning ```const char&``` or '''char``` ?

course5/homework06/string.h
```diff
@@ -0,0 +1,43 @@
...
+
+    String& operator +=(const String& string);
+
+    bool operator == (const String& s2) const;
+    bool operator == (const char * str) const;
```
maciekgajewski, line 26: operator != is missing

course5/homework06/string.cc
```diff
@@ -0,0 +1,120 @@
...
+
+String::String(char * data)
+    :String((const char *)data){}
+
+String::String(const char *data)
+{
+    if (!data)
```
maciekgajewski: This is not really needed.
One of the principles of C_++ is: "don't pay for what you're not using". 
In this case it's OK to just document the function as requiring the pointer not to be null. If it's null, the program will crash. If user suspects that the pointer may be null, their will checking before calling the function.
If user is sure that the pointer is never null, the this 'if' will be redundand.

course5/homework06/string.cc
```diff
@@ -0,0 +1,120 @@
...
+}
+
+std::size_t String::size() const
+{
+    return this->length();
+}
+
+String::~String()
+{
+    this->free();
```
maciekgajewski: Why do you do it?

course5/homework06/string.cc
```diff
@@ -0,0 +1,120 @@
...
+    return this->buffer.get()[index];
+}
+
+const char& String::operator [](size_t index) const
+{
+    return this->buffer.get()[index];
+}
+
+
+void String::free()
```
maciekgajewski: This function is absolutely not needed.

course5/homework06/string.cc
```diff
@@ -0,0 +1,120 @@
...
+{
+    if (this->buffer)
+    {
+        this->buffer.reset();
+        this->buffer = nullptr;
```
maciekgajewski: Even if it would be needed, it's 3 times too big.
1. ```reset()``` releases the memory and sets the pointer value to null
2. assigning ```nullptr``` to ```unique_ptr``` is the same as calling ```reset()```
3. both of the above operation are idempotent and work just fine if the pointer is already null.

course5/homework06/string.cc
```diff
@@ -0,0 +1,120 @@
...
+    auto r = String(str);
+    return r;
+}
+
+String& String::operator +=(const String& s)
+{
+    auto new_size = this->size() + s.size();
+    auto new_buffer = std::make_unique<char[]>(new_size);
+
+    strcat(new_buffer.get(), this->buffer.get());
```
maciekgajewski: Awlays use fully qualified versions of functions. Otherwise you risk using C instead of C++ implementations, and making you code not compatible with future version of C++ compiler.
So: ```std::strcat```

course5/homework06/string.cc
```diff
@@ -0,0 +1,120 @@
...
+    data.free();
+}
+
+std::size_t String::length() const
+{
+    if (buffer)
+    {
+        return std::strlen(this->buffer.get());
+    }
+    throw std::runtime_error("invalid state of an instance. buffer doesn't point to any meaningful value");
```
maciekgajewski: Why not just return 0? What is std::string doing?

course5/homework06/string.cc
```diff
@@ -0,0 +1,120 @@
...
+}
+
+String::~String()
+{
+    this->free();
+}
+
+const char* String::c_str() const
+{
+    return this->buffer.get();
```
maciekgajewski: what will it return if the string is empty?

course5/homework06/string.cc
```diff
@@ -0,0 +1,120 @@
...
+{    
+    this->free();
+    this->buffer = std::move(string.buffer);
+    string.free();
+    return *this;
+}
+
+bool String::operator ==(const String& str) const
+{
+    return strcmp(this->c_str(), str.c_str()) == 0;
```
maciekgajewski: Use ```std::strcmp```, include <cstring> in this file if yuo're using this function.

course5/homework06/string.cc
```diff
@@ -0,0 +1,120 @@
...
+String operator +(const String& s1, const String& s2)
+{
+    auto size = s1.size() + s2.size();
+    auto p = std::make_unique<char[]>(size);
+    char* str = p.get();
+    str = std::strcat(str, s1.c_str());
+    str = std::strcat(str, s2.c_str());
+
+    auto r = String(str);
+    return r;
```
maciekgajewski: This implementation is buggy, and your testing is insufficient if you haven't detected it.
- the buffer is too short
- it would blow-up entirely if any of the strings is empty
- there is unnecessary deep copy at the end, which together with O(N) implementation of size() makes it a slow hog

course5/homework06/string.cc
```diff
@@ -0,0 +1,120 @@
...
+String& String::operator +=(const String& s)
+{
+    auto new_size = this->size() + s.size();
+    auto new_buffer = std::make_unique<char[]>(new_size);
+
+    strcat(new_buffer.get(), this->buffer.get());
+    this->free();
+    strcat(new_buffer.get(), s.c_str());
+    this->buffer = std::move(new_buffer);
+    return *this;
```
maciekgajewski: Similar bugs here like in ```operator+```

course5/homework06/test.h
```diff
@@ -0,0 +1,103 @@
...
+        auto s = T(str);
+        assert(s == "this is string with length 29");
+    }
+
+    void test_move_constructor()
+    {
+        auto str = "this is string with length 29";
+        auto s = T(str);
+        auto s2 = std::move(s);
+        assert (s2 == "this is string with length 29");
```
maciekgajewski, line 26: How do you know it was a move and not copy?

course5/homework06/test.h
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    void test_copy_assignment()
+    {
+        auto s1 = T("blaah");
+        T s2 = s1;
+
+        assert (s2.length() == 5);
+        assert (s1.length() == 5);
+        assert (s1 == s2);
```
maciekgajewski, line 36: test that the copy is deep, ie that changing s1 after the copy doesn't affect s2

course5/homework06/test.h
```diff
@@ -0,0 +1,103 @@
...
+        auto str = "this is string with length 29";
+        auto s = T(str);
+        assert(s.length() == 29);
+    }
+
+    void test_size()
+    {
+        auto str = "this is string with length 29";
+        auto s = T(str);
+        assert(s.size() == 29);
```
maciekgajewski, line 144: Test corner cases:
- that default-constructed string is empty and size() == 0
- test adding empty strign to non-empty string, adding non-empty string to empty string, adding two empty strings together

maciekgajewski (changes requested):
- The behaviour doesn't match std::String, especially for empty/default constructed strings
- I suspect that the idea of how unique_ptr work escaped you. It is supposed to make your code simpler and safer, while you add a lot of code to deal with it.
- I don't think you've tested the program with valgrind, did you?

course5/homework06/test.h
AJIEIIIA: added

course5/homework06/test.h
AJIEIIIA: added

course5/homework06/test.h
AJIEIIIA: doesn't std::move guarantee that it's move?

course5/homework06/string.h
AJIEIIIA: added

course5/homework06/string.h
AJIEIIIA: the same? no?

course5/homework06/string.h
AJIEIIIA: yes, it's const char* now.

course5/homework06/string.cc
AJIEIIIA: ok, removed

course5/homework06/string.cc
AJIEIIIA: thanks for explanation, indeed...

course5/homework06/string.cc
AJIEIIIA: fixed

course5/homework06/string.cc
AJIEIIIA: fixed

course5/homework06/string.cc
AJIEIIIA: fixed, I guess...

course5/homework06/string.cc
AJIEIIIA: missed that one

course5/homework06/string.cc
AJIEIIIA: fixed

course5/homework06/string.cc
AJIEIIIA: fixed

course5/homework06/string.cc
AJIEIIIA: yes, true. Now it returns 0;

course5/homework06/string.cc
AJIEIIIA: this check also used for copy constructor, since it calls this constructor.

course5/homework06/main.cpp
AJIEIIIA: fixed

course5/homework06/string.h
maciekgajewski: Possibly. char is 1 byte, the reference is 8, but I suspect it's all the same on 64 bit machines.

course5/homework06/test.h
maciekgajewski: move only casts normal reference into r-value reference. It's up to the author of the class (you!) to ensure that when rvalue ref (```&&```) is ued, you actually move.
You should test it. One way you can so it is to compare the value returned by ```data()``` or ```c_str()```: s.c_str() before move should return the same as s2.c_str() after move. The same address. This would mean that s2 is using the same buffer s used to use.

PR #153, opened 2017-04-11
Homework 10-1

sophiadavis:
Sorry about the diff. Just the last commit has the iterator stuff.

### DIFF
--- a/course5/homework06/main.cc
+++ b/course5/homework06/main.cc
@@ -288,31 +288,53 @@ int main(int, char**)
 
     std::cout << std::boolalpha;
     auto any_a = std::any_of(iteratorTest.begin(), iteratorTest.end(), isA);
+    auto any_a_const = std::any_of(iteratorTest.cbegin(), iteratorTest.cend(), isA);
     assert (any_a);
+    assert (any_a_const);
 
     bool are_equal = std::equal(iteratorTest.begin(), iteratorTest.end(), iteratorTest2.begin());
+    bool are_equal_const = std::equal(iteratorTest.cbegin(), iteratorTest.cend(), iteratorTest2.cbegin());
     assert (!are_equal);
+    assert (!are_equal_const);
 
     auto max_el = std::max_element(iteratorTest.begin(), iteratorTest.end(), std::greater<char>());
     auto max_el_index = std::distance(iteratorTest.begin(), max_el);
+    auto max_el_const = std::max_element(iteratorTest.cbegin(), iteratorTest.cend(), std::greater<char>());
+    auto max_el_index_const = std::distance(iteratorTest.cbegin(), max_el_const);
     int index_first_space = 4;
     assert (max_el_index == index_first_space);
+    assert (max_el_index_const == index_first_space);
 
     char t = 't';
     int num_t = std::count(iteratorTest.begin(), iteratorTest.end(), t);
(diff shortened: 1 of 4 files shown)


course5/homework06/main.cc
```diff
@@ -0,0 +1,314 @@
...
+
+    String iteratorTest = "this is a test";
+    String iteratorTest2 = "this is another test";
+    for (auto& elem: iteratorTest) {
+        std::cout << elem << std::endl;
+    }
+
+    std::cout << std::boolalpha;
+    auto any_a = std::any_of(iteratorTest.begin(), iteratorTest.end(), isA);
+    std::cout << any_a << std::endl;
```
maciekgajewski: Please ,automate the tests

course5/homework06/simple_string.h
```diff
@@ -0,0 +1,49 @@
...
+    char operator[](const int) const;
+    char& operator[](const int);
+
+    void Print() const;
+    int length() const;
+    const char *c_str() const;
+
+    typedef char * iterator;
+    typedef const char * const_iterator;
+    iterator begin() { return &buffer.get()[0]; }
```
maciekgajewski: Knowing that for any pointer ```p``` and integer ```i```, the following expression are identical:
```p[i]``` === ```*(p + i)```
and
```&(*p)``` === ```p```
And knowing that ```buffer.get()``` is a pointer

How can you simplify this function?

File: course5/homework06/simple_string.h [hunk 3]
```diff
@@ -0,0 +1,49 @@
...
+    iterator end() { return &buffer.get()[l]; }
```
maciekgajewski: I don't think this would work

maciekgajewski (changes requested):
Please, automate your tests, because I don't believe your implementation of end() actually works.

Also - you need a const version of begin and end (+ tests!)

You should also test if you're able to modify the content of your string via iterators.

course5/homework06/simple_string.h
sophiadavis: It seems to be working...? What am I doing wrong?

course5/homework06/simple_string.h
```diff
@@ -0,0 +1,52 @@
...
+    char& operator[](const int);
+
+    void Print() const;
+    int length() const;
+    const char *c_str() const;
+
+    typedef char * iterator;
+    typedef const char * const_iterator;
+    iterator begin() { return buffer.get(); }
+    iterator end() { return buffer.get() + l; }
```
maciekgajewski, line 37: OK. it is l (small-caps L), not 1 (one). So confusing...
See, ```mLength``` or ```length_``` or ```this->l``` would be much easier to read.

Of course it's good.

course5/homework06/simple_string.h
```diff
@@ -0,0 +1,52 @@
...
+    char operator[](const int) const;
+    char& operator[](const int);
+
+    void Print() const;
+    int length() const;
+    const char *c_str() const;
+
+    typedef char * iterator;
+    typedef const char * const_iterator;
+    iterator begin() { return buffer.get(); }
```
maciekgajewski, line 36: You also need const begin and end, otherwise you wouldn't be able to iterate over const object.

maciekgajewski (commented):
Of course it's good; I was confused by combination of bad typeface and your minimalistic variable names.

PR #155, opened 2017-04-12
Homework10.1

jfdsmit:


course5/homework10.1/string.h
```diff
@@ -0,0 +1,44 @@
...
+	bool operator!=(String const & that) const;
+	bool operator!=(char const * that) const;
+	char const * c_str() const;
+	String::iterator begin();
+	String::const_iterator begin() const;
+	String::iterator end();
+	String::const_iterator end() const;
+
+private:
+	std::unique_ptr<char[]> mBuffer = nullptr;
```
maciekgajewski, line 36: this ```= nullptr``` is not needed at all. `uinque pointer's default constructor would take care of that.

maciekgajewski (commented):
Good!

I would put all the one-liners into the header file ,so that compiler can inline them; but otherwise this is the simplest and correct solution.

PR #157, opened 2017-04-12
Homework 10-2 (very, very wip)

sophiadavis:
Current status -- `begin` seems to be working, but `end` was giving me issues because I hadn't defined an `==` operator on the linked lists yet, which became a rabbit hole.

So I spent some more time trying to figure out where memory leaks and errors are coming from 😞


course5/homework10/linked_list.h
```diff
@@ -0,0 +1,221 @@
...
+    LinkedList& operator=(LinkedList&& src)
+    {
+        std::cout << "Move assignment: from " << src.front() << std::endl;
+        data = std::move(src.data);
+        next = std::move(src.next);
+        mSize = src.mSize;
+        return *this;
+    }
+
+    bool operator==(LinkedList& cmp) const
```
sophiadavis, line 80: I'm having trouble with `==` because the compiler insists on calling the binary version.

course5/homework10/linked_list.h
```diff
@@ -0,0 +1,221 @@
...
+        if (next)
+        {
+            LinkedList<T> new_head = std::move(*next.get());
+            data = std::move(new_head.data);
+            next = std::move(new_head.next);
+            mSize--;
+        }
+    }
+
+    void print() const
```
sophiadavis, line 147: This guy causes memory errors down the wazoo. There has got to be a better way to debug this stuff than adding printfs before and after every line, and seeing where valgrind complains?

sophiadavis: And I'd like to move to iterators here, but I'd also like to understand what's going wrong.

course5/homework10/linked_list.h
```diff
@@ -0,0 +1,221 @@
...
+        }
+        bool operator!=(LinkedListIter & cmp)
+        {
+            return !(this == cmp);
+        }
+        T & operator*() {
+            // if (my_list.data)
+                return *mList.data.get();
+            // else
+                // return *nullptr;
```
sophiadavis, line 199: Couldn't figure this case out...

course5/homework10/linked_list.h
```diff
@@ -0,0 +1,221 @@
...
+        }
+    private:
+        LinkedList<T>& mList;
+};
+
+// template<class T>
+// bool operator==(const LinkedList<T>& cmp1, const LinkedList<T>& cmp2)
+// {
+//     std::cout << " binary " << std::endl;
+//     return cmp1.operator==(&cmp2);
```
sophiadavis, line 218: Back to my `==` issues, I was thinking I could also define a public `check_next__nodes_are_equal` method, but I don't really want any idea of the `next` to be public. I'd rather force the compiler to call the unary version.

course5/homework10/linked_list.h
```diff
@@ -0,0 +1,221 @@
...
+template<class T, class Allocator = std::allocator<T>>
+class LinkedList
+{
+public:
+    friend class LinkedListIter<T>;
+    typedef LinkedListIter<T> iterator;
+
+    // Basic assignment constructors
+    LinkedList() = default;
+    LinkedList(const T& val) // TODO T&& val ?
```
sophiadavis, line 21: `push_front` on an empty list doesn't seem to cause memory leaks, but once I use this constructor, I start leaking memory (I think this is it, at least). I don't think I should need to define a destructor (all the linked list contains is some `unique_ptr` -- shouldn't it just work?).

course5/homework10/linked_list.h
```diff
@@ -0,0 +1,221 @@
...
+        }
+        std::cout << std::endl;
+    }
+
+    iterator begin() { return iterator(*this); }
+    iterator end() { return iterator(nullptr); }
+
+private:
+    std::unique_ptr<T> data;
+    std::unique_ptr<LinkedList<T>> next;
```
maciekgajewski, line 178: I think the entire architecture of the list is wrong. I suggest you take a pencil, paper, and draw it.

course5/homework10/linked_list.h
```diff
@@ -0,0 +1,221 @@
...
+            return false;
+        return true;
+    }
+
+    int size() const { return mSize; }
+    T front() const // TODO should this be a pointer? const?
+    {
+        if (data)
+            return *data.get();
+        throw std::range_error("Calling front on empty LinkedList is undefined.");
```
maciekgajewski, line 99: This is not true. Calling ```front``` on your list produces a behaviour that's very well defined: it throws an exception.
If you'd remove the if/throw, than you can call it 'undefined', and you'd also move the burden of checking if the list is not empty to the user. This would prevent the user for "paying for what they don't use"

course5/homework10/linked_list.h
```diff
@@ -0,0 +1,221 @@
...
+        {
+            LinkedList<T> new_node;
+            new_node.data = std::move(this->data);
+            new_node.next = std::move(this->next);
+            data = std::make_unique<T>(args...);
+            next = std::make_unique<LinkedList<T>>(std::move(new_node));
+        }
+        else
+        {
+            data = std::make_unique<T>(args...);
```
maciekgajewski, line 131: You need to use ```std::forward``` here. I'll leave it as an exercise to you to find out why :)

course5/homework10/linked_list.h
```diff
@@ -0,0 +1,221 @@
...
+        }
+        mSize++;
+    }
+
+    void pop_front()
+    {
+        if (next)
```
maciekgajewski, line 138: and else?

course5/homework10/linked_list.h
maciekgajewski: I think you should start by re-designing the entire list,and then we can focus on the mem errors.

course5/homework10/linked_list.h
maciekgajewski: And rightly so. This one doesn't adhere to the canonical interface of the operator.

maciekgajewski (changes requested):
You've made some nice progress since the first PR, but the entire premise of the container is wrong.

I suggest you start over. May be easier than trying to fix this one.

PR #159, opened 2017-04-19
homework 11.1, benchmarking containers

jfdsmit:


course5/homework11.1/homework11.1.cc
```diff
@@ -0,0 +1,64 @@
...
+	{
+		temp.push_back(list.front());
+		list.pop_front();
+	}
+	list.push_front(element);
+	while (!temp.empty())
+	{
+		list.push_front(temp.back());
+		temp.pop_back();
+	}
```
maciekgajewski, line 28: You could just ```push_front``` here. The order in which the integers sit in the container doesn't matter for the outcome. Or does it?

course5/homework11.1/homework11.1.cc
```diff
@@ -0,0 +1,64 @@
...
+}
+
+
+template <typename T>
+static void BM_Container(benchmark::State & state)
+{
+	T container;
+	for (int i = 0; i < 100; i++)
```
maciekgajewski, line 36: Run the benchmark for different container sizes

course5/homework11.1/homework11.1.cc
```diff
@@ -0,0 +1,64 @@
...
+// Results
+// jfdsmit@paranor:~/werk/optiver/c-excercises/course5/homework11.1/build$ make && ./homework11.1
+// [100%] Built target homework11.1
+// Run on (2 X 2400 MHz CPU s)
+// 2017-04-18 22:20:06
+// --------------------------------------------------------------------------
+// Benchmark                                   Time           CPU Iterations
+// --------------------------------------------------------------------------
+// BM_Container<std::vector<int>>           1884 ns       1883 ns     368819
+// BM_Container<std::list<int>>             1083 ns       1082 ns     642249
```
maciekgajewski, line 59: Surprising result! Wonder if the same observation would stand for different container sizes.
Also - if making few allocation between inserting stuff to the list would change the result.

maciekgajewski: Just by looking at the number, it looks like a debug build...

maciekgajewski (changes requested):
Test for different container lengths. Make sure its not a debug build!

PR #160, opened 2017-04-19
homework 11.2

jfdsmit:


course5/homework11.2/homework11.2.cc
```diff
@@ -0,0 +1,57 @@
...
+static void BM_VectorSequential(benchmark::State & state)
+{
+	std::vector<int> v(10000000, 42);
+	while (state.KeepRunning())
+	{
+		int result = 0;
+		for (size_t i = 0; i < v.size(); i++)
+		{
+			size_t index;
+			benchmark::DoNotOptimize(index = i % 2 ? i : i - 0);
```
maciekgajewski, line 14: This doesn't really prevent the optimisation of the expression inside. The function only stops the compiler from throwing the expression away.

course5/homework11.2/homework11.2.cc
```diff
@@ -0,0 +1,57 @@
...
+			result += v[index];
+		}
+	}
+}
```
maciekgajewski, line 18: I think you are underestimating the power of optimizer. This becnhmark simply does nothing.
We'll discuss it one day

maciekgajewski (commented):
The becnhamrk doesn't really do what you think it does. But to verify this, we'd have to look at the generated assembly.

PR #162, opened 2017-04-24
Dani Homework 10.2

vargad:


course5/homework10/linkedlist.hh
```diff
@@ -0,0 +1,209 @@
...
+        --mSize;
+    }
+
+    std::size_t size() const
+    {
+        return mSize;
+    }
+
+private:
+    std::unique_ptr<Node> mHead = nullptr;
```
maciekgajewski: No need to initialize

course5/homework10/linkedlist.hh
```diff
@@ -0,0 +1,209 @@
...
+        ++mSize;
+        Node *tail = mHead.get();
+        auto it = begin_iter;
+        ++it;
+        for (; it != end_iter; ++it)
+        {
+            tail->next = std::make_unique<Node>(nullptr, *it);
+            tail = tail->next.get();
+            ++mSize;
+        }
```
maciekgajewski: I think you could re-write the loop in a much simpler form.

course5/homework10/linkedlist.hh
```diff
@@ -0,0 +1,209 @@
...
+        {
+            return node->data;
+        }
+
+        const ValueType& operator*() const
+        {
+            return node->data;
+        }
+    private:
+        NodeType node;
```
maciekgajewski: It's not clear that ```NoteType``` is a pointer. Using ```NodeType*``` and passing const/non const ```Node``` as template argument would be cleaner.

course5/homework10/homework10.cc
```diff
@@ -0,0 +1,221 @@
...
+            l2.push_front(i);
+        assert(std::is_sorted(l2.begin(), l2.end()));
+    }
+    {
+        std::list<NonCopyAble> l1;
+        LinkedList<NonCopyAble> l2;
+        for (int i=0; i<10; ++i)
+        {
+            l1.emplace_front("Test"+std::to_string(i));
+            l2.emplace_front("Test"+std::to_string(i));
```
maciekgajewski: ```push_front``` should work with non-copyable, movable objects. Test this scenario please.

maciekgajewski (commented):
Good; simple.
push_front wouldn't work for move-only objects; it should.

PR #163, opened 2017-04-25
Homework12

jfdsmit:


course5/homework12/line_sorting.cc
```diff
@@ -0,0 +1,47 @@
...
+
+namespace jds {
+
+std::vector<std::string> ReadLines(std::istream & source)
+{
+	std::vector<std::string> lines;
+	std::string line;
+	while (source)
+	{
+		std::getline(source, line);
```
maciekgajewski: ```std::getline``` returns source; you could use it to compact the code a bit.

course5/homework12/line_sorting.cc
```diff
@@ -0,0 +1,47 @@
...
+		if (bool(source) || line.length())
+			lines.push_back(std::move(line));
+	}
+	return lines;
+}
+
+std::string FindColValue(std::string const & line, char separator, size_t column)
+{
+	size_t pos = 0;
```
maciekgajewski: Consider using Boost.Tokenizer

course5/homework12/line_sorting.cc
```diff
@@ -0,0 +1,47 @@
...
+
+std::vector<std::string> ReadLines(std::istream & source)
+{
+	std::vector<std::string> lines;
+	std::string line;
+	while (source)
+	{
+		std::getline(source, line);
+		if (bool(source) || line.length())
+			lines.push_back(std::move(line));
```
maciekgajewski: Q: why std::move here? What does it change?

maciekgajewski (commented):
Good!

course5/homework12/line_sorting.cc
jfdsmit: According to Compiler Explorer, without `std::move` here the compiler will generate a call to `std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >::push_back(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)` which looks like it will trigger a copy constructor because the argument is of type `std::string const&`. With `std::move` the argument type changes to `std::string &&` to allow a move construction call.

course5/homework12/line_sorting.cc
jfdsmit: Nice

course5/homework12/line_sorting.cc
jfdsmit: Of course. I briefly looked at `strtok()` and quickly discarded it as an option, but should have realized that if there was no tokenizer in the standard library there should have been one in Boost

course5/homework12/line_sorting.cc
maciekgajewski: Correct! You avoid deep copy of the string. But does it improve the overall performance of the function?

PR #165, opened 2017-04-30
Dani Homework 11+extra

vargad:


course5/homework11/homework11.cc
```diff
@@ -0,0 +1,178 @@
...
+        c.push_back(i);
+    if (order == random_read)
+    {
+        std::shuffle(std::begin(c), std::end(c), std::default_random_engine(seed));
+    }
+
+    while(state.KeepRunning())
+    {
+        size_t pos = 0;
+        while (pos<c.size())
```
maciekgajewski, line 162: Does this correctly control the number of iterations?
if it happens so ```pos[0] = size```, then there will be only one iteration of the loop.

course5/homework11/homework11.cc
```diff
@@ -0,0 +1,178 @@
...
+            pos = c[pos];
+    }
+}
+
+static void CustomListSize(benchmark::internal::Benchmark* b)
```
maciekgajewski, line 167: Why not just use Range?

maciekgajewski (commented):
I'm not sure the random benchmark is accurate

PR #166, opened 2017-05-02
Homework12

sophiadavis:


course5/homework12/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+FIND_PACKAGE(Boost COMPONENTS program_options REQUIRED)
+INCLUDE_DIRECTORIES(${Boost_INCLUDE_DIRS})
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -lboost_program_options")
```
maciekgajewski: you don't add libraries by setting compilation flags, but rather by using ```target_link_libraries```.
Setting explicit compilation flags should be avoided, as this breaks portability (GCC vs Clang is usually OK, but think of Visual Studio!).

course5/homework12/main.cc
```diff
@@ -0,0 +1,164 @@
...
+
+    if (vm.count("col")) {
+        column = vm["col"].as<int>();
+    }
+
+    if (vm.count("sep")) {
+        separator = vm["sep"].as<std::string>();
+    }
+    
+    std::unique_ptr<std::vector<std::string>> buffer = std::make_unique<std::vector<std::string>>();
```
maciekgajewski: Why have you decided to create unique_ptr to vector?

course5/homework12/main.cc
```diff
@@ -0,0 +1,164 @@
...
+            std::string msg = std::strerror(errno);
+            throw std::runtime_error("Failed to open file: " + msg);
+        }
+    }
+    else
+    {
+        read(std::cin, *buffer.get());
+    }
+
+    boost::sort(*buffer.get(), sorter(descending, column, separator));
```
maciekgajewski: ```unique_ptr``` has all the needed operators overloaded. In your code you use it just like if it was a raw pointer.
So in this case: ```*buffer```

course5/homework12/main.cc
```diff
@@ -0,0 +1,164 @@
...
+
+    bool operator() (const std::string& s1, const std::string& s2)
+    {
+        std::string to_cmp_1 = s1;
+        std::string to_cmp_2 = s2;
+
+        std::vector<std::string> split_s1;
+        std::vector<std::string> split_s2;
+
+        if (m_column != 0 && m_separator == "")
```
maciekgajewski: This looks like something that should be part of initialisation/validation of sorting parameters, not something you'd do every time when comparing elements.

course5/homework12/main.cc
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+void write(std::ostream& out, const std::vector<std::string>& buffer)
+{
+    for(auto item : buffer)
+        out << item << std::endl;
+}
+
+struct sorter
```
maciekgajewski: Try using lambda!

course5/homework12/main.cc
```diff
@@ -0,0 +1,164 @@
...
+
+        if (m_descending)
+            return to_cmp_1 > to_cmp_2;
+        return to_cmp_1 < to_cmp_2;
+    }
+
+    bool m_descending = false;
+    int m_column = 0;
+    std::string m_separator = "";
+};
```
maciekgajewski: Use lambda instead. Using functor allows you to make some complex things easily; for instance in this case you could think of several interesting performance optimizations. But since you are not doing them, just go for lambda.

maciekgajewski (commented):
Why unique_ptr? Why functor?

PR #168, opened 2017-05-21
Homework12

AJIEIIIA:


course5/homework12/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+project(homework12)
+cmake_minimum_required(VERSION 3.2)
+
+find_package(Boost REQUIRED COMPONENTS program_options)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -lboost_program_options")
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_BUILD_TYPE Debug)
```
maciekgajewski, line 7: This setting doesn't belong int the file.
When configuring the build from command -line, it should be the user decision which mode they want.
 Tools, like Visual Studio or QtCreator, will generate multiple configurations and allow the user the switch between them easily.

course5/homework12/comparers.h
```diff
@@ -0,0 +1,83 @@
...
+};
+
+std::unique_ptr<AComparer> CreateComparer(bool numeric, bool keyed, unsigned int column, char separator);
+
+class BasicComparer : public AComparer
+{
+public:
+    bool operator() (const std::string& s1, const std::string& s2) const override
+    {
+        return std::strcmp(s1.c_str(), s2.c_str()) < 0;
```
maciekgajewski, line 24: Why not simply ```return s1 < s2;``` ??

course5/homework12/comparers.h
```diff
@@ -0,0 +1,83 @@
...
+    {
+        try
+        {
+            auto d1 = std::stod(s1);
+            auto d2 = std::stod(s2);
+            return d1 < d2;
+        }
+        catch(std::invalid_argument&)
+        {
+            return std::strcmp(s1.c_str(), s2.c_str()) < 0;
```
maciekgajewski, line 41: and here

course5/homework12/comparers.h
```diff
@@ -0,0 +1,83 @@
...
+    }
+};
+template <typename TComparer>
+struct ColumnComparer: public AComparer
+{
+public:
+    ColumnComparer(int colNumber, char separator)
+                   :colNumber(colNumber), separator(separator), comparer(TComparer())
+    {
+        if (this->colNumber <= 0)
```
maciekgajewski, line 52: Why don't validate params when params are obtained? Then you could get away with a simple ```assert``` here.

course5/homework12/comparers.h
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+private:
+    const std::string FindColumnValue(const std::string& s) const
+    {
+        const char * separators = {&this->separator};
+        boost::char_separator<char> sep {separators};
+        boost::tokenizer<boost::char_separator<char>> tokenizer {s, sep};
+        auto cur = tokenizer.begin();
+        int ix = 1;
```
maciekgajewski, line 70: you can put the declarartion into the for:
```for(int ix = 1; ix < colNumber && ...)```

File: course5/homework12/homework12.cc [hunk 6]
```diff
@@ -0,0 +1,138 @@
...
+{
+    po::variables_map vm;
+    po::store(po::parse_command_line(argc, argv, desc), vm);
+    po::notify(vm);
+
+    return vm;
+}
+}
+
+std::vector<std::string> ReadLines(std::istream&  file)
```
maciekgajewski, line 45: Mark all your functions except of ```main``` as ```static```. This will inform the compiler that no linker symbols have to be generated for the, as they are not used outside of the module and don't have to participate in linking. 
This will also give the compiler more optimization opportunities.

course5/homework12/homework12.cc
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    return lines;
+}
+
+void WriteLines(std::ostream& stream, std::vector<std::string>& lines)
+{
+    for (auto it = lines.begin(); it != lines.end(); ++it)
+    {
+        stream << it->c_str() << std::endl;
```
maciekgajewski, line 62: What is going o here? Why manual iteration in the loop? Why c_str??

course5/homework12/homework12.cc
```diff
@@ -0,0 +1,138 @@
...
+        }
+        output = &tmp_out;
+    }
+
+    auto comparer = homework12::CreateComparer(numeric, keyed, colIndex, separator);
+
+    auto lines = ReadLines(*input);
+    std::sort(lines.begin(), lines.end(), [&comparer](std::string a, std::string b){return comparer->operator ()(a, b);});
+    if (reverse)
+        std::reverse(lines.begin(), lines.end());
```
maciekgajewski, line 133: This may be expensive, in case of large files. You could do it smarter, by either: reversing the order of sorting, or by printing the output in reverse order.

course5/homework12/comparers.h
```diff
@@ -0,0 +1,83 @@
...
+#include <stdexcept>
+#include <boost/tokenizer.hpp>
+#include <memory>
+
+namespace homework12{
+
+class AComparer
+{
+public:
+    bool virtual operator() (const std::string& s1, const std::string& s2) const = 0;
```
maciekgajewski, line 14: You are not using this class as a functor; instead, you wrap it in a lambda. Why using an overloaded function call operator then? I think simple function, like ```virtual bool Compare(a, b) const``` would be clearer.

course5/homework12/comparers.cc
```diff
@@ -0,0 +1,26 @@
...
+    }
+    else
+    {
+        if (numeric)
+            return std::make_unique<NumericComparer>();
+        else
+            return std::make_unique<BasicComparer>();
+
+    }
+}
```
maciekgajewski, line 23: Nice pattern: interface + factory building template-customized types. Very nice, but used in a rather inefficient way.

You build the object once, based on one input (the params), but then you pay the cost of calling virtual function O(N log(N)) times!

If, instead, your interface contained the entire call to ```std::sort```, then you'd call the virtual function only once, and then all the comparisons inside of std::sort would be inlined. Consider this!

course5/homework12/tests.cpp
```diff
@@ -0,0 +1,38 @@
...
+    assert (target2("-0999", "8"));
+}
+
+int main(int, char**)
+{
+    NumericComparerTest();
+    ColumnComparerTest();
+
+    return 0;
+}
```
maciekgajewski, line 38: Awesome! I like tests

maciekgajewski (commented):
Nice, but I see possibilities for few easy, but significant performance improvements.

PR #170, opened 2017-06-14
Ack project wip

sophiadavis:
Here's what I have so far -- it's still kinda messy and all-in-one-file and doesn't handle erroneous input well etc. But it mostly works in the basic case, I think.

### DIFF
--- a/course5/ack/main.cc
+++ b/course5/ack/main.cc
@@ -72,21 +72,27 @@ int main(int argc, char** argv)
         return 1;
     }
 
-    std::vector<std::string> search_paths;
+    std::vector<std::string> input_search_paths;
+    std::vector<bfs::directory_entry> paths_to_process;
     if (vm.count("search_paths"))
     {
-        search_paths = vm["search_paths"].as<std::vector<std::string>>();
+        input_search_paths = vm["search_paths"].as<std::vector<std::string>>();
     }
     else
     {
-        search_paths = {"."};
+        input_search_paths = {"."};
     }
-    for (std::string path : search_paths)
+    for (std::string path : input_search_paths)
     {
         for (bfs::directory_entry& bfs_path : bfs::recursive_directory_iterator(path))
         {
-            std::cout << "---> searching: " << bfs_path.path() << std::endl;
+            paths_to_process.push_back(bfs_path);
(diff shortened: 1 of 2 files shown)


course5/ack/main.cc
```diff
@@ -0,0 +1,186 @@
...
+            write(std::cout, bfs_path.path().string());
+            write(std::cout, match_buffer);
+            write(std::cout, "\n");
+        }
+    }
+
+    if (cow)
+    {
+        std::vector<std::string> cow_pic = {
+"         ______",
```
maciekgajewski, line 175: :)
Consider raw string literal here

course5/ack/main.cc
```diff
@@ -0,0 +1,186 @@
...
+Add command-line parameters to select different regex styles, case in/sensibility, coloring of output etc.
+Get inspired by actual “ack”.
+
+Next steps:
+* Search stdin
+* actually write to file if desired
+* colors
+***/
+
+void find_matching_lines(const boost::filesystem::directory_entry& fs_entry, const std::regex& pattern, std::vector<std::string>& match_buffer)
```
maciekgajewski, line 23: This is correct code, I don't have too many c++ comments, so I'll comment on the design:
1. the function does 2 things: opens file and searches for patter in opened file. You would achieve greater reusability if you split it. This way you could search for patterns in ifstream that's not a disk file, for instance STDIO.
2. The canonical way of returning multiple results from an algortihm function is to consume output iterator. See std::copy or std::transform for instance

course5/ack/main.cc
```diff
@@ -0,0 +1,186 @@
...
+        {
+            write(std::cout, bfs_path.path().string());
+            write(std::cout, match_buffer);
+            write(std::cout, "\n");
+        }
+    }
+
+    if (cow)
+    {
+        std::vector<std::string> cow_pic = {
```
maciekgajewski, line 174: This should be ```const char[] cow_pic = ...```, and the data would be static in static area.
When using vector, the data still sits in static area, but then, in the vector's constructor, a heap buffer is allocated and data copied.

maciekgajewski (commented):
Alright!

PR #171, opened 2017-08-24
Homework1

bodorkos-adam:
Added a better main draw loop. Note this code is NOT how I think this should be done in a nice way, but that would involve far too much code for now :)

### DIFF
--- a/course6/toyapp/CMakeLists.txt
+++ b/course6/toyapp/CMakeLists.txt
@@ -1,10 +1,11 @@
 project(toyapp)
 cmake_minimum_required(VERSION 3.0)
 
+link_directories(C:/lib/SDL2/lib)
 add_subdirectory(displaylib)
 
 
-add_executable(toyapp
+add_executable(toyapp WIN32
     main.cc)
 
 set_property(TARGET toyapp PROPERTY CXX_STANDARD 14)
--- a/course6/toyapp/displaylib/rgb.h
+++ b/course6/toyapp/displaylib/rgb.h
@@ -1,0 +1,17 @@
+#pragma once
+
+#include <cstdint>
+
+namespace Display {
+
+struct Rgb
+{
+	std::uint8_t r = 0;
+	std::uint8_t g = 0;
+	std::uint8_t b = 0;
+};
+
+constexpr Rgb YELLOW{255, 255, 0};
+constexpr Rgb BLUE{0, 0, 128};
+
+}
--- a/course6/toyapp/displaylib/sdl.h
+++ b/course6/toyapp/displaylib/sdl.h
@@ -1,0 +1,4 @@
+#pragma once
+
+#define SDL_MAIN_HANDLED
+#include <SDL2/SDL.h>
--- a/course6/toyapp/displaylib/surface.cc
+++ b/course6/toyapp/displaylib/surface.cc
@@ -15,23 +15,42 @@ Surface::~Surface()
 		SDL_FreeSurface(mSurface);
 }
 
-void Surface::Clear(std::uint8_t r, std::uint8_t g, std::uint8_t b)
+void Surface::Clear(Rgb color)
 {
 	assert(mSurface);
 
 	for(int y = 0; y < mSurface->h; y++)
 		for(int x = 0; x < mSurface->w; x++)
-			SetPixel(x, y, r, g, b);
+			SetPixel(Pixel{x, y}, color);
 }
 
-void Surface::SetPixel(int x, int y, std::uint8_t r, std::uint8_t g, std::uint8_t b)
+void Surface::SetPixel(Pixel position, Rgb color)
 {
 	assert(mSurface);
 
-	std::uint32_t pixelValue = (std::uint32_t(r)<<16) + (std::uint32_t(g)<<8) + b;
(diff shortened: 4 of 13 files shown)


course6/toyapp/displaylib/rgb.h
```diff
@@ -0,0 +1,17 @@
...
+namespace Display {
+
+struct Rgb
+{
+	std::uint8_t r = 0;
+	std::uint8_t g = 0;
+	std::uint8_t b = 0;
+};
+
+constexpr Rgb YELLOW{255, 255, 0};
```
maciekgajewski, line 14: You should just use ```const``` here. ```constexpr``` has other uses (eve thou it would work here)

course6/toyapp/displaylib/sdl.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
+
+#define SDL_MAIN_HANDLED
```
maciekgajewski, line 3: This looks like a hack. Why is it needed?
I think it's needed because you use "main" without parameters. This is OK with "vanilla" main, which gets special treatment by the compiler, but once you include SDL_main.h, SDL will define a macro called "main", which resolves to ```SDL_main(int, char**)```. This maycause compilation failure if you use it without ```(int, char**)``` params.

course6/toyapp/displaylib/surface.h
```diff
@@ -14,9 +15,10 @@ class Surface : private SDLObject
 	Surface(int w, int h);
 	~Surface();
 
-	void Clear(std::uint8_t r, std::uint8_t g, std::uint8_t b);
-	void SetPixel(int x, int y, uint8_t r, uint8_t g, uint8_t b);
-
+	void Clear(Rgb color);
+	void SetPixel(Pixel position, Rgb color);
+	void DrawVertex(Vector3D vertex, Rgb color);
+	void DrawCube(Cube& cube, Rgb color);
```
maciekgajewski, line 21: Do you think is a good design? Should Surface know about Cube or 3D Vertex?
I think it should be Cube that knows about the Surface. This way you can expand your catalogue of shapes, without having to teach Surface about every one of them.

Consider moving the methods to they respective objects:

```
Vector3D::Draw(Surface& surf);
Cube::Draw(Surface& surf);
```

course6/toyapp/displaylib/vector.cc
```diff
@@ -0,0 +1,13 @@
+#include "vector.h"
+
+namespace Display {
+
+Vector2D Vector3D::project() const
```
maciekgajewski, line 5: The naming of the function is not consistent with the project's style, which uses CamelCasing

course6/toyapp/CMakeLists.txt
```diff
@@ -1,10 +1,11 @@
 project(toyapp)
 cmake_minimum_required(VERSION 3.0)
 
+link_directories(C:/lib/SDL2/lib)
```
maciekgajewski, line 4: I've found an elegant solution for this:
```
if (SDL_ROOT)
   include_directories(${SDL_ROOT}/include)
   link_directories(${SDL_ROOT}/lib)
endif()
```
This way you only have to set ```SDL_ROOT ``` variable in your project's config, which you can do in the CMake GUI, or by setting it in command-line, and you don't need to hardcode the path in your CMakeLists.txt

maciekgajewski (changes requested):
Please, copy your toyapp to toyapp-adam

course6/toyapp/displaylib/surface.cc
```diff
@@ -15,23 +15,42 @@ Surface::~Surface()
...
+	std::uint32_t* pixelAddr = static_cast<std::uint32_t*>(mSurface->pixels) + offset;
 	*pixelAddr = pixelValue;
 }
 
+void Surface::DrawVertex(Vector3D vertex, Rgb color)
+{
+	auto point2D = vertex.project();
+	Pixel halfScreen{mSurface->w / 2, mSurface->h / 2};
+	Pixel pixel {
+		static_cast<int>(halfScreen.x + point2D.x * halfScreen.x),
```
mtux, line 43: Am I right that this is to move cube to center and scale it up?
I've tried drawing without this (with just dividing x and y by z) and my cube became to tiny that almost wasn't visible. This looks to resolve that issue of mine.

maciekgajewski: yes, this changes the coordinates from physical pixels to (-1.0 - +1.0)

course6/toyapp/displaylib/sdl.h
bodorkos-adam: Thanks, I will check out SDL main!

course6/toyapp/displaylib/surface.h
bodorkos-adam: You're completely right.

course6/toyapp/displaylib/vector.cc
bodorkos-adam: My bad, I'm using lowerCamelCase in my home project and did this out of habit.

bodorkos-adam:
I will create a new PR with the second homework before the session next week.

PR #172, opened 2017-08-30
Homework-1

mtux:


course6/toyapp-mehrdad/displaylib/types.h
```diff
@@ -0,0 +1,47 @@
...
+	int mHeight;
+};
+
+struct Color
+{
+	uint8_t mRed;
+	uint8_t mGreen;
+	uint8_t mBlue;
+
+	uint32_t GetPixelValue() const { return (uint32_t(mRed)<<16) + (uint32_t(mGreen)<<8) + mBlue; }
```
maciekgajewski, line 44: This format is specific to the surface we're using, and IMO the conversion should be there

maciekgajewski (approved):
nice!

course6/toyapp-mehrdad/displaylib/types.h
mtux: Right. Will move it there.

PR #173, opened 2017-08-30
Homework toyapp

vivikelapoutre:


course6/toyapp-vivike/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -std=c++14")
+
+project(toyapp)
+cmake_minimum_required(VERSION 3.0)
+
+# Ugly way of providing the path on Windows
```
maciekgajewski, line 6: You can remove this commented-out garbage, if you don't use it.

course6/toyapp-vivike/displaylib/objects.h
```diff
@@ -0,0 +1,33 @@
...
+	void move(int xmove, int ymove);
+};
+
+struct Vector3D
+{
+	int x;
+	int y;
+	int z;
+
+	Vector2D project2D();
```
maciekgajewski, line 30: To project from 3d to 2d, you may require some extra context, like camera position or aspect ratio.
The way we do it it's super-simple, but we wouldn't like to lock the simplicity in the design.

I believe that projecting 2D -> 3D should be a job of separate object. We can call it Camera or Surface3D

course6/toyapp-vivike/main.cc
```diff
@@ -0,0 +1,46 @@
...
+
+
+int main(int, char**)
+{
+	std::cout << "Hello" << std::endl;
+
+	Display::Window win("Hello", 10, 10, 800, 600);
+	Display::Surface surf(200, 150);
+
+	Display::Color blue;
```
maciekgajewski, line 18: Or:
```
Display::Color blue{0, 0, 255};
```

maciekgajewski (approved):
The solution is incomplete.
3d->2d conversion is missing, and you're only displaying the lone pixel instead of some 3D stuff.

PR #175, opened 2017-08-30
Minimalistic Raytracing

a-moreno-a:


a-moreno-a:
Different approach to transform objects in view space to image space, using raycasting to perform the projection

course6/toyapp_ariel/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+project(toyapp)
+cmake_minimum_required(VERSION 3.0)
+
+# Ugly way of providing the path on Windows
```
maciekgajewski, line 4: If you don't need this, you should remove it. No need to keep commented-out corpses of code around.

course6/toyapp_ariel/displaylib/float3.h
```diff
@@ -0,0 +1,29 @@
+#ifndef FLOAT3_H
+#define FLOAT3_H
+
+#include <algorithm>
+#include <math.h>
```
maciekgajewski, line 5: Don't use C headers (like "math.h"), use their C++ counterparts ("cmath")

course6/toyapp_ariel/displaylib/float3.h
```diff
@@ -0,0 +1,29 @@
+#ifndef FLOAT3_H
+#define FLOAT3_H
+
+#include <algorithm>
```
maciekgajewski, line 4: You don't use the headers in this, so they shouldn't be here. 
Include files only where you use them, otherwise you only slow down the compilation.

course6/toyapp_ariel/displaylib/float3.h
```diff
@@ -0,0 +1,29 @@
...
+#include <math.h>
+
+class Float3
```
maciekgajewski, line 7: I don't like the name, it means nothing. If I would encounter the name while reading the code, it wouldn't be immediately obvious to me that this is vector.

course6/toyapp_ariel/displaylib/float3.h
```diff
@@ -0,0 +1,29 @@
...
+    Float3 Normalize() const;
+    float Dot(const Float3& f) const;
+    float Length() const;
+    Float3 Clamp() const;
+    Float3 operator / (const float& f) const;
+    Float3 operator - (const Float3& f) const;
+    Float3 operator + (const Float3& f) const;
+    Float3 operator * (const float& f) const;
+    Float3 operator * (const Float3& f) const;
+    static float Clamp(const float& x)
```
maciekgajewski, line 23: This is ```std::clamp```. If it's in the standard library; don't reinvent it!

course6/toyapp_ariel/displaylib/float3.h
```diff
@@ -0,0 +1,29 @@
...
+{
+public:
+    float x, y ,z;
+    Float3();
+    Float3(float f);
+    Float3(const float& x, const float& y, const float& z);
+    Float3 Normalize() const;
+    float Dot(const Float3& f) const;
+    float Length() const;
+    Float3 Clamp() const;
```
maciekgajewski, line 17: Clamp to what? More obvious name or at least a comment would be useful.

course6/toyapp_ariel/displaylib/float3.h
```diff
@@ -0,0 +1,29 @@
...
+#include <algorithm>
+#include <math.h>
+
+class Float3
+{
+public:
+    float x, y ,z;
+    Float3();
+    Float3(float f);
+    Float3(const float& x, const float& y, const float& z);
```
maciekgajewski, line 13: The interface is inconsistent: here use pass paramas by const ref, above by copy.
The const reference buys you nothing in this case, only increases the typing required.

course6/toyapp_ariel/displaylib/float3.cc
```diff
@@ -0,0 +1,61 @@
...
+{
+
+}
+
+Float3 Float3::Clamp() const
+{
+    return Float3(Float3::Clamp(x), Float3::Clamp(y), Float3::Clamp(z));
+}
+
+float Float3::Length() const
```
maciekgajewski, line 23: For performance reasons, it is better to put one-liners in the header file. This way compiler can inline them when used in another source files.
Inlining is the most important optimisation; it acts as an enabler for many other optimisations.

course6/toyapp_ariel/displaylib/material.h
```diff
@@ -0,0 +1,24 @@
...
+enum MaterialType
+{
+    Diffuse,
+    Specular
+};
+
+class Material
+{
+public:    
+    Material(){}
```
maciekgajewski, line 15: So, if you default-construct the object, what the value of ```matType``` will be?

course6/toyapp_ariel/displaylib/renderingengine.cc
```diff
@@ -0,0 +1,94 @@
...
+
+RenderingEngine::RenderingEngine(const int &imageSize) : imageSize(imageSize)
+{
+    image = std::vector<Float3>(imageSize * imageSize);
+}
+
+Ray RenderingEngine::CreatePrimaryRay(int x, int y)
+{
+    int fov = 90;
+    float Px = (2 * ((x + 0.5) / imageSize) - 1) * std::tan(fov / 2 * M_PI / 180);
```
maciekgajewski, line 11: You wouldn't have to use ```std::tan``` if you'd assume flat projection, ie just use pixel's coordinates to construct the ray

course6/toyapp_ariel/displaylib/renderingengine.h
```diff
@@ -0,0 +1,31 @@
+#ifndef RENDERINGENGINE_H
```
maciekgajewski, line 1: consider using ```#pragma once``` instead

course6/toyapp_ariel/displaylib/renderingengine.h
```diff
@@ -0,0 +1,31 @@
...
+#define RENDERINGENGINE_H
+
+#include <math.h>
```
maciekgajewski, line 4: ```<cmath>```

course6/toyapp_ariel/main.cc
```diff
@@ -0,0 +1,47 @@
...
+        {
+            Float3 currentPixel = image[i * IMAGE_SIZE + j];
+            surf.SetPixel(i, j, currentPixel.x * 255, currentPixel.y * 255, currentPixel.z * 255); // red pixel at 10x10
+        }
+    }
+
+    win.Display(surf);
+    Display::Delay(100000);
+
+	Display::Delay(1000);
```
maciekgajewski, line 44: why 2 delays in a row?

course6/toyapp_ariel/main.cc
```diff
@@ -0,0 +1,47 @@
...
+
+    Display::Window win("Hello", 10, 10, IMAGE_SIZE, IMAGE_SIZE);
+    Display::Surface surf(IMAGE_SIZE, IMAGE_SIZE);
+
+    for(int i = 0; i < IMAGE_SIZE; i++)
+    {
+        for(int j = 0; j < IMAGE_SIZE; j++)
+        {
+            Float3 currentPixel = image[i * IMAGE_SIZE + j];
+            surf.SetPixel(i, j, currentPixel.x * 255, currentPixel.y * 255, currentPixel.z * 255); // red pixel at 10x10
```
maciekgajewski, line 37: I don't the comment reflects what the code is doing :)

course6/toyapp_ariel/displaylib/sphere.cc
```diff
@@ -0,0 +1,22 @@
...
+{
+    Float3 op = o - r.GetOrigin();
+    float t, eps = 1e-4f, b = op.Dot(r.GetDirection()), det = b * b - op.Dot(op) + rad * rad;
+    if(det < 0)
+    {
+        return IntersectionInfo();
+    }
+    else
+    {
+        det = sqrt(det);
```
maciekgajewski, line 18: Use ```std::sqrt```

course6/toyapp_ariel/displaylib/renderingengine.h
```diff
@@ -0,0 +1,31 @@
...
+#include <vector>
+#include <memory>
+
+#include "float3.h"
+#include "sphere.h"
+#include "primitive.h"
+
+class RenderingEngine
```
maciekgajewski, line 12: I have doubts about the design.

This class serves 2 purposes: it performs rendering by ray-tracing, and stores the scene information.

IMO,  cleaner design would be to have 2 separate classes:
* one that stores the scene info: primitives, lights etc
* another one that uses the data to render the objects

This way you can have different renderers, and apply them to the same scene.

I would also prefer if the rendering engine render directly to the surface, instead of storing data in intermediate buffer. This creates a code duplication; the Surface class _is_ already the intermediate buffer.

course6/toyapp_ariel/displaylib/renderingengine.h
```diff
@@ -0,0 +1,31 @@
...
+public:
+    RenderingEngine(const int& imageSize);
+    void AddPrimitive(std::shared_ptr<Primitive> primitive);
+    void SetLightPosition(const Float3& lightPosition);
+    const std::vector<Float3> Render();
+    Float3 RayTrace(int depth, const Ray& ray);
+private:
+    Float3 lightPosition;
+    std::vector<Float3> image;
+    std::vector<std::shared_ptr<Primitive>> primitives;
```
maciekgajewski, line 23: I think you over-designed it here a bit...

course6/toyapp_ariel/displaylib/intersectioninfo.h
```diff
@@ -0,0 +1,22 @@
...
+#include "material.h"
+
+class IntersectionInfo
+{
+public:
+    IntersectionInfo(){t = 0;}
+    IntersectionInfo(const float& t, const Float3& n, const Material& material);
+    Material& GetMaterial(){return material;}
+    Float3 GetNormal() {return n;}
+    float GetT() {return t;}
```
maciekgajewski, line 14: What is "T"?

course6/toyapp_ariel/displaylib/intersectioninfo.h
```diff
@@ -0,0 +1,22 @@
...
+    void SetT(float v){ t = v;}
+private:
+    float t;
+    Float3 n;
```
maciekgajewski, line 18: The code would be more readable if you used more descriptive names.

course6/toyapp_ariel/displaylib/intersectioninfo.h
```diff
@@ -0,0 +1,22 @@
+#ifndef INTERSECTIONINFO_H
+#define INTERSECTIONINFO_H
+
+#include "float3.h"
+#include "material.h"
+
+class IntersectionInfo
+{
+public:
+    IntersectionInfo(){t = 0;}
```
maciekgajewski, line 10: Do you really need the default constructor?

course6/toyapp_ariel/displaylib/primitive.cc
```diff
@@ -0,0 +1,6 @@
+#include "primitive.h"
```
maciekgajewski, line 1: This file is not compiled, not added to CMake.
If it were, it wouldn't compile, as the method it defines is not declared anywhere.

course6/toyapp_ariel/displaylib/sphere.h
```diff
@@ -0,0 +1,21 @@
...
+#include "primitive.h"
+#include "material.h"
+
+class Sphere : public Primitive
+{
+public:
+    Sphere(const Float3& o, const float& rad, const Material& material);
+    IntersectionInfo Intersect(const Ray& r);
+private:
+    Float3 o;
```
maciekgajewski, line 15: what is ```o```? Bad naming!

course6/toyapp_ariel/displaylib/sphere.h
```diff
@@ -0,0 +1,21 @@
...
+    float rad;
+    Float3 c;
```
maciekgajewski, line 17: what is ```c```? I don't see it being used anywhere

maciekgajewski (approved):
Cool effect bro, but...
The code sloppy. Surprising or information-deprived names, some artefacts left here and there, files not even added to the project etc.

It's better to write less but nicer code. Aim for readability and simplicity.

course6/toyapp_ariel/CMakeLists.txt
a-moreno-a: Removed

course6/toyapp_ariel/displaylib/float3.h
a-moreno-a: Fixed

course6/toyapp_ariel/displaylib/float3.h
a-moreno-a: Fixed

course6/toyapp_ariel/displaylib/float3.h
a-moreno-a: The built-in type for a 3 components vector in HSLS is "float3" that's why the motivation of using this name

course6/toyapp_ariel/displaylib/float3.h
a-moreno-a: Its true that I get nothing from keeping float values inmutable in this case, removed the constant reference

course6/toyapp_ariel/displaylib/primitive.cc
a-moreno-a: Removed

course6/toyapp_ariel/displaylib/float3.h
a-moreno-a: This seems to be something new from c++17, dont really know how to make it work

a-moreno-a: Changed the name in hte meantime

course6/toyapp_ariel/displaylib/float3.cc
a-moreno-a: oh, after the comment I read a bit and found out that implementing functions inside the class declarations doesnt break the "one time definition" rule, did not know about it, changed

course6/toyapp_ariel/displaylib/material.h
a-moreno-a: Fixed

course6/toyapp_ariel/displaylib/renderingengine.h
a-moreno-a: Done

course6/toyapp_ariel/displaylib/renderingengine.h
a-moreno-a: Fixed

PR #176, opened 2017-08-31
Initial version

dhuyst:
Work in progress

Creating a Draw::Renderer object (main.cc:18) throws the following run-time error:

libGL error: dlopen /usr/lib64/dri/swrast_dri.so failed (/usr/lib64/dri/swrast_dri.so: cannot open shared object file: No such file or directory)
libGL error: unable to load driver: swrast_dri.so
libGL error: failed to load driver: swrast
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  154 (GLX)
  Minor opcode of failed request:  3 (X_GLXCreateContext)
  Value in failed request:  0x0
  Serial number of failed request:  87
  Current serial number in output stream:  88

### DIFF
--- a/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
+++ b/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
@@ -8,6 +8,7 @@ add_library(drawlib
     drawer.cc drawer.h
     projector.cc projector.h
 	vector3d.cc vector3d.h
+    matrix.cc matrix.h
 )
 
 set_property(TARGET drawlib PROPERTY CXX_STANDARD 14)
--- a/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
+++ b/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
@@ -1,6 +1,10 @@
 add_library(drawlib
 	primitives.cc primitives.h
 	index.h
+    rgb.h
+    colors.h
+    renderer.cc renderer.h
+    render_object.cc render_object.h
 )
 
 set_property(TARGET drawlib PROPERTY CXX_STANDARD 14)
--- a/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
+++ b/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
@@ -1,6 +1,6 @@
 add_library(drawlib
 	primitives.cc primitives.h
-	index.h
+	index.cc index.h
     rgb.h
     colors.h
     renderer.cc renderer.h
--- a/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
+++ b/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
@@ -6,6 +6,7 @@ add_library(drawlib
     renderer.cc renderer.h
     render_object.cc render_object.h
     drawer.cc drawer.h
+    projector.cc projector.h
 )
 
 set_property(TARGET drawlib PROPERTY CXX_STANDARD 14)
--- a/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
+++ b/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
@@ -5,6 +5,7 @@ add_library(drawlib
     colors.h
     renderer.cc renderer.h
     render_object.cc render_object.h
+    drawer.cc drawer.h
 )
 
 set_property(TARGET drawlib PROPERTY CXX_STANDARD 14)
--- a/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
+++ b/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
@@ -1,0 +1,9 @@
+add_library(drawlib
+	primitives.cc primitives.h
+	index.h
+)
+
+set_property(TARGET drawlib PROPERTY CXX_STANDARD 14)
+target_link_libraries(drawlib displaylib)
+
+target_include_directories(drawlib INTERFACE ..)
--- a/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
+++ b/course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
@@ -7,6 +7,7 @@ add_library(drawlib
     render_object.cc render_object.h
     drawer.cc drawer.h
     projector.cc projector.h
+	vector3d.cc vector3d.h
 )
 
 set_property(TARGET drawlib PROPERTY CXX_STANDARD 14)
--- a/course6/toyapp_thomasdhuys/drawlib/drawer.h
+++ b/course6/toyapp_thomasdhuys/drawlib/drawer.h
@@ -1,0 +1,17 @@
+#pragma once
+#include <vector>
+#include "primitives.h"
+#include "render_object.h"
+
+namespace Draw {
+
+class Drawer {
+public:
+	static RenderObject Draw();
+
+private:
+	static std::vector<Index> GetSquareIndices();
+	static RenderObject DrawSquarePoints();
+};
+
+}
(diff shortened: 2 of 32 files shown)


course6/toyapp_thomasdhuys/drawlib/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+add_library(drawlib
+	primitives.cc primitives.h
+	index.h
+    rgb.h
```
maciekgajewski: inconsistent indentation

course6/toyapp_thomasdhuys/main.cc
```diff
@@ -0,0 +1,37 @@
...
+
+
+
+
+	for(int x = 0; x < 100; x++)
+	{
+//		surf.Clear(0, 0, 255); // blue background
+//		surf.SetPixel(10+x, 10, 255, 0, 0); // red pixel at 10x10
+//		win.Display(surf);
+//		Display::Delay(50);
```
maciekgajewski: don't keep commented-out code around

course6/toyapp_thomasdhuys/drawlib/renderer.cc
```diff
@@ -0,0 +1,25 @@
+#include "renderer.h"
+
+namespace Draw {
+
+void Renderer::Render(RenderObject ro)
+{
+	for (const auto& prim_ptr : *ro.GetData())
+	{
+
```
maciekgajewski: this doesn't seem to do much

course6/toyapp_thomasdhuys/drawlib/renderer.h
```diff
@@ -0,0 +1,26 @@
...
+class Renderer {
+public:
+	Renderer();
+	void Render(RenderObject ro);
+private:
+	static Display::Window DefaultWindow();
+	static Display::Surface DefaultSurface();
+	Display::Window window;
+	Display::Surface surface;
+};
```
maciekgajewski: I don't see much purpose in this class. It's just a one function, that even doesn't do anything.

course6/toyapp_thomasdhuys/drawlib/render_object.h
```diff
@@ -0,0 +1,26 @@
...
+#include <memory>
+#include <string>
+
+#include "primitives.h"
+
+namespace Draw {
+
+using namespace Draw::Primitives;
+
+class RenderObject {
```
maciekgajewski, line 13: * The name of the class is very generic and doesn't really reveal it's purpose. Maybe "Scene" or "World" would be better?
* The class doesn't add much value. You can just provide a function that iterates over the container and calls ToString on every element.

course6/toyapp_thomasdhuys/drawlib/index.h
```diff
@@ -0,0 +1,33 @@
...
+#include <cstdint>
+#include <memory>
+
+#include "rgb.h"
+#include "colors.h"
+
+namespace Draw {
+
+struct Index;
+using XCoord = int;
```
maciekgajewski: Surprising type name, especially that you use it also to store Y and Z coordinates.

course6/toyapp_thomasdhuys/drawlib/index.h
```diff
@@ -0,0 +1,33 @@
...
+using IndexP = std::shared_ptr<Index>;
+
+// Treat Index as a memory intensive object
+struct Index {
```
maciekgajewski: I think "Vertex" would be a better name

course6/toyapp_thomasdhuys/drawlib/projector.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "index.h"
+#include "rgb.h"
+
+namespace Draw {
+
+struct Index2D
```
maciekgajewski: And this, i think, would be more descriptive as "Pixel"

course6/toyapp_thomasdhuys/drawlib/primitives.h
```diff
@@ -0,0 +1,43 @@
...
+};
+
+class Triangle: public Primitive {
+public:
+	Triangle(IndexP& idx0, IndexP& idx1, IndexP& idx2);
+	std::string ToString() override;
+private:
+	IndexP idx0;
+	IndexP idx1;
+	IndexP idx2;
```
maciekgajewski: Why do you store pointers to points instead of points themselves?!

course6/toyapp_thomasdhuys/drawlib/index.h
```diff
@@ -0,0 +1,33 @@
...
+#include "rgb.h"
+#include "colors.h"
+
+namespace Draw {
+
+struct Index;
+using XCoord = int;
+using IndexP = std::shared_ptr<Index>;
+
+// Treat Index as a memory intensive object
```
maciekgajewski: What do you mean by "memory intensive"?

course6/toyapp_thomasdhuys/drawlib/drawer.h
```diff
@@ -0,0 +1,17 @@
...
+namespace Draw {
+
+class Drawer {
+public:
+	static RenderObject Draw();
+
+private:
+	static std::vector<IndexP> GetSquareIndices();
+	static RenderObject DrawSquarePoints();
+};
```
maciekgajewski: I don't see a point in this object as well. It represents no concept, it's just a collection of functions

course6/toyapp_thomasdhuys/drawlib/drawer.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+#include <vector>
+#include "primitives.h"
+#include "render_object.h"
+
+namespace Draw {
+
+class Drawer {
+public:
+	static RenderObject Draw();
```
maciekgajewski, line 10: The function doesn't really "draw" anything, it just returns a collection of vertices.

maciekgajewski (changes requested):
This is so unnecessarily complex! So many classes and types, and shared pointers and collections - just to draw a simple cube.

Try again; make it 10x simpler.

Also:
- keep an eye on your indentation; don't mix spaces and tabs
- don't leave commented-out code around
- try not to leave all these debug prints everywhere. It's easier to use a debugger than to generate verbose log

course6/toyapp_thomasdhuys/drawlib/renderer.h
dhuyst: It's supposed to draw Lines and Triangles from their respective primitives, but I didn't get there yet.

course6/toyapp_thomasdhuys/drawlib/render_object.h
dhuyst: - Scene sounds like a better name indeed.
- Meant to use this object as an interface between the "Drawer" which generates the scene and the "Renderer" which puts the scene on the screen

course6/toyapp_thomasdhuys/drawlib/index.h
dhuyst: Indeed, bad naming. Renamed to VCoord (Vertex Coordinate)

course6/toyapp_thomasdhuys/drawlib/index.h
dhuyst: renamed

course6/toyapp_thomasdhuys/drawlib/projector.h
dhuyst: renamed

course6/toyapp_thomasdhuys/drawlib/primitives.h
dhuyst: Want to mimic opengl implementation. Primitives containing Indices which point to Vertices.

maciekgajewski:
I've looked at your recent code. 
1. You don't call Window::Display
2. You abuse the Displaylib; I'm not sure if copying Window and Surface is implemented correctly. IT is possible that the underlying objects are destroyed while the objects are copied

PR #177, opened 2017-09-05
Adam Homework 2

bodorkos-adam:
* Move the mouse to rotate camera horizontally. I know I can't simply use the Y coordinate in the Euler angle rotation as that will introduce roll, so this is something to fix later. Also, the camera moves along the global X-Y axes, not relative to itself, but this should be easy to fix. I did not want to introduce a Quaternion this time, but might consider it later.
* Use Arrow keys and PageUp / PageDown to move around. Press End to exit. For some twisted reason I had trouble binding the Escape key (scancode) to do this - my app instantly quit ?!
* Don't go too close to the cubes. I warned you!
* The Keyboard class exists 1) to wrap SDL, and 2) to have the ability to tell if the key has just been pressed / released in this Update. Although I don't actually use this yet.
* I rely on Vsync to handle timing. For now.
* I believe goto is not inherently evil.

Until I can think of a "better" way for this project, I simply made the transforms public. Also, something should probably encapsulate Keyboard and Mouse.


course6/toyapp-adam/displaylib/surface.cc
```diff
@@ -0,0 +1,131 @@
...
+	return {
+		static_cast<Pixel::Coordinate>(mHalfDimensions.x + vector2D.x * mHalfDimensions.x),
+		static_cast<Pixel::Coordinate>(mHalfDimensions.y + vector2D.y * mHalfDimensions.y),
+	};
+}
+
+Vector2D Surface3D::Project(Vector3D vector) const
+{
+	const auto clipSpace = camera.GetProjectionMatrix() * Vector4D(vector);
+	return {-clipSpace.x / clipSpace.w, clipSpace.y / clipSpace.w};
```
bodorkos-adam, line 128: I'm not sure why this negation is needed here. But without it, cubes that I positioned in -X were displayed on the right side. Maybe my camera is looking the wrong way - but the left-right movement seems fine. I need to investigate, but I probably have a +/- kind of mistake somewhere.

course6/toyapp-adam/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
+
+project(toyapp)
+cmake_minimum_required(VERSION 3.0)
+
+set(SDL_ROOT "" CACHE STRING "SDL2 Library root path")
+
+include_directories(${SDL_ROOT}/include)
```
maciekgajewski, line 8: This should be only called when SDL_ROOT is set. IT only needs to be set if you have it in a non-standard location

course6/toyapp-adam/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+
+add_subdirectory(displaylib)
+add_subdirectory(math)
+add_subdirectory(util)
+
+add_executable(toyapp WIN32 main.cc)
+set_property(TARGET toyapp PROPERTY CXX_STANDARD 14)
+
+target_link_libraries(toyapp
+#	mingw32                    # HACK, @todo REMOVE and FIX!!
```
maciekgajewski, line 20: make it conditional. CMake should know (ie set appropriate variable) if you're suing mingw.

course6/toyapp-adam/displaylib/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+add_library(displaylib
+	camera.cc
```
maciekgajewski, line 2: Add headers and other files to your project as well. CMake will add them to generated project files, use them to track dependencies etc

course6/toyapp-adam/displaylib/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
...
+	cube.cc
+	scene.cc
+	sdlobject.cc
+	surface.cc
+	window.cc)
+
+set_property(TARGET displaylib PROPERTY CXX_STANDARD 14)
```
maciekgajewski, line 9: Can you set this globally instead of per-target?

course6/toyapp-adam/math/matrix.cc
```diff
@@ -0,0 +1,135 @@
...
+
+namespace {
+	constexpr auto PI = std::atan(1.0) * 4.0;
+}
+
+Matrix44 Matrix44::Zero()
+{
+	Matrix44 m;
+
+	for(unsigned i = 0; i < 16; ++i)
```
maciekgajewski, line 15: Use range-based or a "fill" algorithm. Simple code, less opportunities for bugs:
```for(float& f: m.mMatrix) f = 0.0f;```
or
```std::fil(std::begin(m.Matrix), std::end(m.mMatrix), 0.0f);```
or
```boost::fill(m.mMatrix, 0.0f);```

File: course6/toyapp-adam/math/matrix.h [hunk 7]
```diff
@@ -0,0 +1,34 @@
...
+	static Matrix44 Rotation(Vector3D yawPitchRoll);
+	static Matrix44 Scale(float size);
+	static Matrix44 Perspective(float fieldOfViewDegrees, float aspectRatio);
+
+	Matrix44 operator*(const Matrix44& rhs) const;
+	Vector3D operator*(const Vector3D& rhs) const;
+	Vector4D operator*(const Vector4D& rhs) const;
+private:
+	// this is a row-major matrix
+	static constexpr auto TRANSLATE_X = 3;
```
maciekgajewski, line 22: If you only use them inside of matrix.cc, then there is no point defining the in the header. Just define them as constants on the module lever in matrix.cc.
They are not part of the public API.

course6/toyapp-adam/math/matrix.cc
```diff
@@ -0,0 +1,135 @@
...
+
+	const float ch = std::cos(yawPitchRoll.x);
+	const float sh = std::sin(yawPitchRoll.x);
+	const float cp = std::cos(yawPitchRoll.y);
+	const float sp = std::sin(yawPitchRoll.y);
+	const float cb = std::cos(yawPitchRoll.z);
+	const float sb = std::sin(yawPitchRoll.z);
+
+	const float shsp = sh * sp;
+	const float chsp = ch * sp;
```
maciekgajewski, line 56: 1. You don't need to make them const
2. It's OK to precalculate them for readability, but you need to be aware that it's not required from the performance point of view.

course6/toyapp-adam/math/matrix.cc
```diff
@@ -0,0 +1,135 @@
...
+	m.mMatrix[11] = +1.0f;
+	m.mMatrix[14] = -1.0f;
+
+	return m;
+}
+
+Matrix44 Matrix44::operator*(const Matrix44& rhs) const
+{
+	constexpr unsigned ROWS = 4;
+	constexpr unsigned COLS = 4;
```
maciekgajewski, line 98: I think this is a stylistic overkill; you could just use literal 4 in the code.
The constexprs have no real meaning in this case either.

course6/toyapp-adam/math/transform.h
```diff
@@ -0,0 +1,26 @@
...
+	void Grow(float delta) { mScale += delta; }
+	void Rotate(Vector3D yawPitchRoll) { mOrientation += yawPitchRoll; }
+
+	Matrix44 CreateMatrix() const;
+	Matrix44 CreateInverseMatrix() const;
+private:
+	Vector3D mPosition;
+	Vector3D mOrientation;
+	float mScale = 1.0f;
+};
```
maciekgajewski, line 24: Do you really need this class? This could simply be the Matrix

course6/toyapp-adam/math/vector.h
```diff
@@ -0,0 +1,40 @@
...
+	Vector3D& operator-=(const Vector3D& rhs);
+	Vector3D& operator*=(float scalar);
+};
+
+struct Vector4D
+{
+	float x;
+	float y;
+	float z;
+	float w;
```
maciekgajewski, line 31: why do you not initilize them, while other vectors do initialize members?

course6/toyapp-adam/math/vector.cc
```diff
@@ -0,0 +1,49 @@
...
+{
+	return {x + rhs.x, y + rhs.y, z + rhs.z};
+}
+
+Vector3D Vector3D::operator-(const Vector3D& rhs) const
+{
+	return {x - rhs.x, y - rhs.y, z - rhs.z};
+}
+
+Vector3D Vector3D::operator*(float scalar) const
```
maciekgajewski, line 20: For performance and readability reasons, it's better to keep the one-lines in the class definition.

course6/toyapp-adam/util/clock.cc
```diff
@@ -0,0 +1,16 @@
...
+
+Clock::Clock()
+{
+	mStartTime = SDL_GetPerformanceCounter();
+	mInverseFrequency = 1.0 / static_cast<double>(SDL_GetPerformanceFrequency());
+}
+
+double Clock::GetTotalElapsedSeconds() const
+{
+	return (SDL_GetPerformanceCounter() - mStartTime) * mInverseFrequency;
```
maciekgajewski, line 13: This is correct, but I'd rather you use ```std::high_resolution_clock```

course6/toyapp-adam/util/keyboard.cc
```diff
@@ -0,0 +1,68 @@
...
+	assert(keyCode < NUM_KEYCODES);
+
+	mKeys[keyCode].up = 1;
+	mKeys[keyCode].down = 0;
+	mKeys[keyCode].released = 1;
+}
+
+void Keyboard::ClearChanges()
+{
+	for(KeyCode i = 0; i < NUM_KEYCODES; ++i)
```
maciekgajewski, line 29: Use range-based for

course6/toyapp-adam/util/keyboard.h
```diff
@@ -0,0 +1,38 @@
...
+	void Release(KeyCode keyCode);
+	void ClearChanges();
+	bool IsUp(KeyCode keyCode) const;
+	bool IsDown(KeyCode keyCode) const;
+	bool IsPressed(KeyCode keyCode) const;
+	bool IsReleased(KeyCode keyCode) const;
+private:
+	struct KeyState
+	{
+		std::uint8_t up : 1;
```
maciekgajewski, line 23: I don't get it. What's he difference between ```up``` and ```released``` and ```down``` and ```pressed```?
By the look of it, you could just store one boolean per key: ```isPressed```?

course6/toyapp-adam/util/keyboard.h
```diff
@@ -0,0 +1,38 @@
...
+{
+public:
+	using KeyCode = SDL_Keycode;
+
+	void Press(KeyCode keyCode);
+	void Release(KeyCode keyCode);
+	void ClearChanges();
+	bool IsUp(KeyCode keyCode) const;
+	bool IsDown(KeyCode keyCode) const;
+	bool IsPressed(KeyCode keyCode) const;
```
maciekgajewski, line 18: What's the difference between ```Down``` and ```Pressed```.
This is API is a gross violation of the principle of the least surprise.

course6/toyapp-adam/util/mouse.h
```diff
@@ -0,0 +1,30 @@
...
+	Pixel GetCursorPosition() const { return mCursorPosition; }
+	Pixel GetCursorPositionDelta() const { return mCursorPositionDelta; }
+
+	void AddDeltaPosition(Pixel delta)
+	{
+		mCursorPositionDelta.x += delta.x;
+		mCursorPositionDelta.y += delta.y;
+	}
+
+	void RecordPosition()
```
maciekgajewski, line 19: This API seems to be very bug-prone. You have 2, order-dependent state modifiers. Seems complex for sucha  simple thing.

course6/toyapp-adam/displaylib/surface.h
```diff
@@ -0,0 +1,53 @@
...
+
+	void Clear(Color color);
+	void SetPixel(Pixel position, Color color);
+	void DrawLine(Pixel start, Pixel end, Color color);
+private:
+	SDL_Surface* mSurface;
+};
+
+
+class Surface3D
```
maciekgajewski, line 33: I think this belongs to a separate file

course6/toyapp-adam/displaylib/surface.cc
```diff
@@ -0,0 +1,131 @@
...
+void Surface3D::DrawLine(Vector3D start, Vector3D end, Color color)
+{
+	mSurface.DrawLine(GetPixel(start), GetPixel(end), color);
+}
+
+Pixel Surface3D::GetPixel(Vector3D point) const
+{
+	const auto vector2D = Project(point);
+	return {
+		static_cast<Pixel::Coordinate>(mHalfDimensions.x + vector2D.x * mHalfDimensions.x),
```
maciekgajewski, line 120: Maybe it should be Surface2D's responsibility to convert from logical Vector2D to physical pixels?

course6/toyapp-adam/displaylib/scene.cc
```diff
@@ -0,0 +1,67 @@
...
+	if(mKeyboard.IsDown(SDL_SCANCODE_PAGEDOWN))
+		cameraMove.y = -camMovementSpeed;
+	if(mKeyboard.IsDown(SDL_SCANCODE_DOWN))
+		cameraMove.z = -camMovementSpeed;
+	if(mKeyboard.IsDown(SDL_SCANCODE_UP))
+		cameraMove.z = +camMovementSpeed;
+
+	mSurface.camera.transform.Move(cameraMove);
+
+	const Pixel screenDimensions = mSurface.GetDimensions();
```
maciekgajewski, line 42: HA! maybe te mouse class should give pointer coordinates in logical units, os that you don't have to do the physical/logical conversion all over your codebase

course6/toyapp-adam/displaylib/surface.cc
a-moreno-a: Have not seen your code really so maybe this is not relevant, but if you trasform the camera (i.e. Rotate eye vector) you should do the same for the up vector, inverted x axis is usually because the is camera upside down

course6/toyapp-adam/main.cc
```diff
@@ -0,0 +1,59 @@
...
+				case SDL_MOUSEMOTION:
+					mouse.AddDeltaPosition({event.motion.xrel, event.motion.yrel});
+					break;
+				case SDL_QUIT:
+					goto quit;
+			}
+		}
+
+		if(keyboard.IsPressed(SDL_SCANCODE_END))
+			goto quit;
```
maciekgajewski, line 45: ```goto``` is not inherently evil. I think your mixing two different quotes attributed to Dijkstra. It was premature optimisation that was evil. In Dijsktra's quotes, gotos where simply the indication of programmers stupidity :)

Use boolean variable to control ```while``` instead.

maciekgajewski (approved):
1. Nice, but could be simpler. I think the transformation class is not needed. So are the Mouse and Keyboard classes- you can send the events straight to Scene (or whatever interested receivers you have.
2. Don't use C-style for loops
3. You seem to apply some misguided optimisations: meaningless constexprs, precalculations etc. Don't
4. CMake - SDL_ROOT should not be mandatory. IT's only needed in exceptional cases

course6/toyapp-adam/util/keyboard.h
```diff
@@ -0,0 +1,38 @@
...
+	{
+		std::uint8_t up : 1;
+		std::uint8_t down : 1;
+		std::uint8_t pressed : 1;
+		std::uint8_t released : 1;
+
+		KeyState() : up{0}, down{0}, pressed{0}, released{0} {}
+	};
+
+	static_assert(sizeof(KeyState) == 1);
```
maciekgajewski, line 31: This doesn't compile in C++14! ```static_assert``` requires 2 arguments

course6/toyapp-adam/displaylib/scene.cc
```diff
@@ -0,0 +1,67 @@
+#include "scene.h"
+
+#include "surface.h"
+#include "util/keyboard.h"
+#include "util/mouse.h"
+#include <cmath>
+
+namespace Display {
+
+Scene::Scene(Keyboard& keyboard, Mouse& mouse, Surface3D& surface)
```
maciekgajewski, line 10: This _is_ your app. This shouldn't be part of the display library

course6/toyapp-adam/math/matrix.h
bodorkos-adam: You're completely right, no idea why I put them in the header.

course6/toyapp-adam/math/vector.h
bodorkos-adam: I initialize them in the constructor. I thought it's a bad idea to mix this kind of initialization with initializer lists if there are multiple constructors defined. Is it not?

course6/toyapp-adam/util/keyboard.h
bodorkos-adam: Difference is that a key might be in a pressed (down) state, but then we can't know whether it has just been pressed in the current Update.

Since our project is a simulation running in a loop, in my opinion it's perfectly valid for the codebase to be aware that something could be related to the current Update only (pressed, released), or a state of the simulation in general (down, up).

Indeed, input could be event-based only instead. That's a design decision and for now I chose polling key states instead. I think having _both_ events and checking state would be the optimal solution.

maciekgajewski (commented):
The mouse events are completely wrong on Linux. Didn't figured it out yet.

course6/toyapp-adam/main.cc
bodorkos-adam: I wasn't trying to quote anyone :) That's quite a strong statement which I wasn't expecting from you. Please don't tell me I'm stupid.

The bool variable would need to be compared to false in every single iteration. Why must I have that? Will the compiler optimize it away? Maybe. But I'm not a fan of premature pessimization either. It took no effort to write it this way.

Actually, what I wanted is to break the SDL event poll loop from within the switch, and figured I don't need stupid additional flags, since I just want to break the hell out of there.

course6/toyapp-adam/displaylib/scene.cc
bodorkos-adam: You mean for example a Vector2D in range [-1 to +1]? Sounds good.

course6/toyapp-adam/util/mouse.h
bodorkos-adam: Sorry I'm not sure where you're getting at. Why should this Mouse worry about what order its methods are called, why is this a bad sign? Reason for this method is the same as keyboard's pressed and down states. The state must be cleared between simulation Updates.

I accumulate SDL mouse move events, since as far as I know more than one can happen between updates. Unless SDL does some kind of magic.

course6/toyapp-adam/util/keyboard.cc
bodorkos-adam: Neat, I did not know it works for plain arrays, thanks!

course6/toyapp-adam/util/clock.cc
bodorkos-adam: Right, that's a good idea!

course6/toyapp-adam/math/matrix.cc
bodorkos-adam: I was under the impression it's preferable to use constexpr if possible. It's not something I'm very sure about though, so I can stop doing this. :)

course6/toyapp-adam/math/matrix.cc
bodorkos-adam: I'm aware it's not required, maybe it's a bad habit of mine. However I do like to make stuff const by default (not function arguments though), but yeah maybe for simple things I shouldn't.

course6/toyapp-adam/math/matrix.cc
bodorkos-adam: Neat, I did not know it works for plain arrays, thanks!

course6/toyapp-adam/displaylib/surface.cc
bodorkos-adam: I don't even have an up vector yet, so this might be it!

course6/toyapp-adam/math/transform.h
bodorkos-adam: I think I do, it encapsulates movement, orientation and scale. Which do not belong IN a matrix, but are used to construct matrices. What if I want to e.g. scale an object which is already rotated and positioned somewhere?

course6/toyapp-adam/displaylib/surface.cc
bodorkos-adam: You're right, it should be!

course6/toyapp-adam/util/keyboard.h
bodorkos-adam: Weird, it compiled for me. I'll add the second argument though.

course6/toyapp-adam/displaylib/scene.cc
bodorkos-adam: You're right.

PR #179, opened 2017-09-07
Fixes after code review on Homework3 PR

bodorkos-adam:
Not everything is fixed, I'll have to move some things to other namespace and/or separate source file.


course6/toyapp-adam/CMakeLists.txt
```diff
@@ -1,23 +1,30 @@
...
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 project(toyapp)
 cmake_minimum_required(VERSION 3.0)
 
 set(SDL_ROOT "" CACHE STRING "SDL2 Library root path")
 
-include_directories(${SDL_ROOT}/include)
+if(NOT SDL_ROOT STREQUAL "")
```
maciekgajewski, line 10: Simple ```if (SDL_ROOT)``` would suffice

course6/toyapp-adam/math/matrix.cc
```diff
@@ -6,14 +6,20 @@ namespace Math {
 
 namespace {
 	constexpr auto PI = std::atan(1.0) * 4.0;
+	constexpr auto TRANSLATE_X = 3;
```
maciekgajewski, line 9: constexpr doesn't mean anything in this case. Just use ```const```

course6/toyapp-adam/util/clock.h
```diff
@@ -1,20 +1,23 @@
...
 public:
-	Clock();
+	Clock()
+	:	mStartTime(HiResClock::now())
+	{}
 
-	double GetTotalElapsedSeconds() const;
+	double GetTotalElapsedSeconds() const
+	{
+		return std::chrono::duration<double>(HiResClock::now() - mStartTime).count();
```
maciekgajewski, line 17: I don't think that count will returns seconds. It will return "ticks", some abstract unit.
To get seconds, just divide the duration by 1 second:
```
using namespace std::literals;
return (HiResClock::now() - mStartTime) / 1s;
```
```

File: course6/toyapp-adam/util/keyboard.cc [hunk 4]
```diff
@@ -26,10 +26,10 @@ void Keyboard::Release(KeyCode keyCode)
 
 void Keyboard::ClearChanges()
 {
-	for(KeyCode i = 0; i < NUM_KEYCODES; ++i)
+	for(auto& key: mKeys)
 	{
-		mKeys[i].pressed = 0;
-		mKeys[i].released = 0;
+		key.pressed = 0;
+		key.released = 0;
```
maciekgajewski, line 32: They can't have the same value at the same time, right? if one is 0, then the other one must 1. Which makes one of them redundant. Which suggests you only need a simple bool.

Apart fro that, I think you should just remove this class entirely.

maciekgajewski (approved):
Better!

course6/toyapp-adam/util/keyboard.cc
bodorkos-adam: They _can_ have the same value at the same time. Between logic updates, the user could have both pressed and released the key very quickly, and we don't want to miss that. The simulation could, if it wants, do X thing when key is pressed, Y when released. In this case I want it to do both.

course6/toyapp-adam/math/matrix.cc
bodorkos-adam: But it does for PI, right?

course6/toyapp-adam/util/clock.h
bodorkos-adam: That doesn't work, the returned value is always a whole number.
It should be okay to use duration.count: http://en.cppreference.com/w/cpp/chrono/duration/duration_cast

PR #180, opened 2017-09-13
Introduce toyapp_mikhails 

Mikhail-SHA-256:
Fixed compiler / linker error in original sources.
Implement method to Draw cube.

Covers meeting01 and partly meeting02-03

### DIFF
--- a/course6/toyapp_mikhail/.gitignore
+++ b/course6/toyapp_mikhail/.gitignore
@@ -1,0 +1,2 @@
+build
+.vs
--- a/course6/toyapp_mikhail/CMakeLists.txt
+++ b/course6/toyapp_mikhail/CMakeLists.txt
@@ -1,0 +1,19 @@
+project(toyapp)
+cmake_minimum_required(VERSION 3.0)
+
+# Ugly way of providing the path on Windows
+#include_directories(D:/SDL)
+#LINK_DIRECTORIES(D:/SDL/lib/win64)
+
+add_subdirectory(displaylib)
+
+
+add_executable(toyapp WIN32
+	main.cc)
+
+set_property(TARGET toyapp PROPERTY CXX_STANDARD 14)
+
+target_link_libraries(toyapp
+	displaylib
+	SDL2main
+	)
--- a/course6/toyapp_mikhail/displaylib/surface.h
+++ b/course6/toyapp_mikhail/displaylib/surface.h
@@ -1,0 +1,60 @@
+#pragma once
+
+#include "sdlobject.h"
+
+#include <SDL2/SDL.h>
+
+#include <cstdint>
+
+namespace Display {
+
+class Surface : private SDLObject
+{
+public:
+	Surface(int w, int h);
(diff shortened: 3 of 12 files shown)


course6/toyapp_mikhail/.gitignore
```diff
@@ -0,0 +1,2 @@
+build
```
maciekgajewski, line 1: It's better to store your build entirely outside of the source dir. This way
a) you can have multiple of them, with different configs
b) you can delete easily when you are running out of disk space
c) last but not least: you're not polluting your source directory

course6/toyapp_mikhail/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+project(toyapp)
+cmake_minimum_required(VERSION 3.0)
+
+# Ugly way of providing the path on Windows
+#include_directories(D:/SDL)
+#LINK_DIRECTORIES(D:/SDL/lib/win64)
```
maciekgajewski, line 6: "love me or leave me" 
don't leave commented-out code.

course6/toyapp_mikhail/displaylib/surface.h
```diff
@@ -0,0 +1,100 @@
...
+		double x, y;
+	};
+
+	class Vector3D
+	{
+	public:
+		double x, y, z;
+	};
+
+	class Matrix
```
maciekgajewski, line 23: Put the classes in separate files please

maciekgajewski (approved):
It's OK for an introduction.
Once the code grows, keep each class in a separate file.

course6/toyapp_mikhail/.gitignore
Mikhail-SHA-256: 1.  added to gitignore for c)
2. I have multiple of them inside build folder
3. I delete entire build folder usually, then do 
 `  mkdir build/build-debug`
 `  cd build/build-debug`
 `  CMake ./../.. `
 `  MSBuild ALL_TRAGETS`
4. Sure I can use folder completely outside of whole git repo but it will be more burden with folders, as my sources not in root of git repo.
  instead of:
`  CMake ./../my-repo`
  it will be
`  CMake ./../../../c-excersises/cource6/toyapp_mikhail`
I will create mycmake.bat to make it easier

course6/toyapp_mikhail/CMakeLists.txt
Mikhail-SHA-256: Was not actually reading / editing these files. That's what I copied from toyapp generic folder.
I plan to review and modify all CMakeList.txt in next pull request.

Will it work for Linux if I add these includes for windows? Currently I don't even need it because I installed SDL2 on system level. Is it OK?

course6/toyapp_mikhail/displaylib/surface.h
Mikhail-SHA-256: Sure! Will be done

PR #181, opened 2017-09-15
Added triangle drawing

bodorkos-adam:
* now drawing cubes with triangles (with crappy algorithm)
* fixed some TODO things left over from previous PRs
* did not fix camera look issues yet. I think it's upside down, so now when you press it actually goes right. will fix


course6/toyapp-adam/displaylib/color.h
```diff
@@ -11,6 +11,24 @@ struct Color
 	Channel r;
 	Channel g;
 	Channel b;
+
+	Color operator*(float intensity)
```
maciekgajewski, line 15: Should this ber const?

course6/toyapp-adam/displaylib/sdlobject.cc
```diff
@@ -13,7 +13,7 @@ SDLObject::SDLObject()
 {
 	if(gOjectCounter++ == 0)
 	{
-		int res = SDL_Init(SDL_INIT_VIDEO);
+		auto res = SDL_Init(SDL_INIT_VIDEO);
```
maciekgajewski, line 16: what was wrong with int?

course6/toyapp-adam/displaylib/surface_2d.cc
```diff
@@ -0,0 +1,137 @@
...
+			if(error < 0)
+			{
+				y += ystep;
+				error += dx;
+			}
+		}
+	}
+}
+
+// slow, naive algorithm. it has overdraw when triangles share edges!
```
maciekgajewski, line 91: That's fine. Don't optimise at this stage.

course6/toyapp-adam/displaylib/surface_2d.cc
```diff
@@ -0,0 +1,137 @@
...
+void Surface2D::DrawTriangle(Pixel a, Pixel b, Pixel c, Color color)
```
maciekgajewski, line 92: You don't need to fill the triangle. It's OK to just draw the edges.

course6/toyapp-adam/math/vector.h
```diff
@@ -42,6 +71,28 @@ struct Vector3D
 		z *= scalar;
 		return *this;
 	}
+
+	Vector3D Normalized() const
+	{
+		auto inverseLength = 1.0f / Length();
```
maciekgajewski, line 77: what if Length is 0?

maciekgajewski (approved):
pretty good!

course6/toyapp-adam/displaylib/sdlobject.cc
bodorkos-adam: you're right, nothing. I just went through the codebase a couple times and opted for consistency with the usage of auto

course6/toyapp-adam/displaylib/color.h
bodorkos-adam: oops, yes!

course6/toyapp-adam/math/vector.h
bodorkos-adam: Good question. Probably would be best to return the vector unchanged, imho.

PR #182, opened 2017-09-18
Homework

asmirnov911:
Without triangles for now.


course6/toyapp_alexs/.vscode/tasks.json
```diff
@@ -0,0 +1,25 @@
+{
```
maciekgajewski: Don't put this into git. This is your private config that reflects your private, local environment.
Add .vscode to your _private_ gitignore file!

course6/toyapp_alexs/displaylib/surface3d.cc
```diff
@@ -0,0 +1,42 @@
...
+#include "vector2d.h"
+#include "vector3d.h"
+
+#include <cassert>
+
+namespace Display {
+
+Surface3D::Surface3D(Surface& surface) : mSurface(surface)
+{
+	mSurface = surface;
```
maciekgajewski: what's going on in here?

course6/toyapp_alexs/displaylib/surface3d.h
```diff
@@ -0,0 +1,27 @@
...
+namespace Display {
+
+class Surface3D 
+{
+public:
+	Surface3D(Surface& surface);
+
+	void Clear(Color color);
+	// TODO ask where it is better to use references and where copy ???
+	void SetPixel(const Vector3D& vector, Color color);
```
maciekgajewski:
```
// TODO answer: pass const ref if copy is expensive. Pass by copy if you actually need the copy (but then yout make the copy inside of a function).
// TODO: so in general: pass by reference if copying reference may be cheaper than copying the object
```

course6/toyapp_alexs/displaylib/transformation3d.cc
```diff
@@ -0,0 +1,75 @@
+#include <cmath>
```
maciekgajewski: Order you headers from the mosty specific. local ones, to the ost generic ones.

So always put the std headers at the end.

course6/toyapp_alexs/displaylib/transformation3d.cc
```diff
@@ -0,0 +1,75 @@
...
+
+#include "vector3d.h"
+#include "transformation3d.h"
+
+namespace Display {
+
+Move3D::Move3D(Vector3D move) : Transformation3D()
```
maciekgajewski: "move" is a confusing name. Consider "translation"

course6/toyapp_alexs/displaylib/transformation3d.cc
```diff
@@ -0,0 +1,75 @@
...
+Vector3D Rotate3D::Apply(Vector3D vertex) const
+{
+    return ApplyZ(ApplyY(ApplyX(vertex)));
+}
+
+Vector3D Rotate3D::ApplyX(Vector3D vertex) const
+{
+    return {
+        vertex.x,
+        cos(mThetaX) * vertex.y - sin(mThetaX) * vertex.z,
```
maciekgajewski: Use qualified functions: ```std::cos```, ```std::sin``` etc

course6/toyapp_alexs/displaylib/transformation3d.h
```diff
@@ -0,0 +1,44 @@
...
+    Vector3D Apply(Vector3D vertex) const;
+private:
+    double mThetaX;
+    double mThetaY;
+    double mThetaZ;
+
+    Vector3D ApplyX(Vector3D vertex) const;
+    Vector3D ApplyY(Vector3D vertex) const;
+    Vector3D ApplyZ(Vector3D vertex) const;
+};
```
maciekgajewski: The entire idea behind transformation matrices is that you don't need to implement every operation separately.
You just construct a matrix representing the transformation, and you multiply vector by the matrix.
You could also combine transformation by multiplying matrices.

maciekgajewski (changes requested):
Quite good, but the transformations should be implemented in terms of transformation matrix.

Also, please, don't put your private config into the repo; I can't merge it like this.

course6/toyapp_alexs/displaylib/transformation3d.cc
asmirnov911: ok, we usually do it vice versa.

course6/toyapp_alexs/.vscode/tasks.json
asmirnov911: will be removed

course6/toyapp_alexs/displaylib/surface3d.cc
asmirnov911: removed

course6/toyapp_alexs/displaylib/surface3d.h
asmirnov911: Thank you for the answer!

course6/toyapp_alexs/displaylib/transformation3d.cc
asmirnov911: what's the difference?

asmirnov911: done

course6/toyapp_alexs/displaylib/transformation3d.h
asmirnov911: will do this a bit later

course6/toyapp_alexs/displaylib/transformation3d.cc
asmirnov911: Well, it moves the object. Translation sounds confusing to me, let's discuss during the meeting.

asmirnov911:
The last commit can be reviewed separately. First, change is self contained. Second, I changed IDE and all the formatting changed in the previous commit.

PR #184, opened 2017-09-20
Implemented class Str,basic operations for string

a-moreno-a:
Implemented basic operator (for example: [], +, +=, = ), constructors(copy and overloads), destructor, created some simple tests in main.cpp to show all operators, ran "valgrind --leak-check=yes -v" and allocations and deallocations match.

*No really sure if std::ostream &operator<<(std::ostream &os, const Str& str) should be inline to avoid redefinition issues


course6/string_ariel/str.h
```diff
@@ -0,0 +1,193 @@
...
+        start = mid = end = nullptr;
+    }
+
+    Str(const char* s)
+    {
+        int size = std::strlen(s);
+        start = new char[size];
+        end = start;
+
+        while (*s != '\0')
```
bodorkos-adam: std::strcpy?

course6/string_ariel/str.h
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    ~Str()
+    {
+        Dispose();
+    }
+
+    char operator[](int i) const
+    {
+        return *(start + i);
```
bodorkos-adam: start[i]

course6/string_ariel/str.h
```diff
@@ -0,0 +1,193 @@
...
+
+        for (int i = 0; i < size2; i++)
+        {
+            *end++ = str2[i];
+        }
+
+        mid = end;
+    }
+
+    friend void Swap(Str& str1, Str& str2)
```
bodorkos-adam: This is usually done differently to enable argument dependent lookup: [https://stackoverflow.com/questions/28130671/how-does-using-stdswap-enable-adl](https://stackoverflow.com/questions/28130671/how-does-using-stdswap-enable-adl)

course6/string_ariel/str.h
a-moreno-a: That’s right

course6/string_ariel/str.h
a-moreno-a: Yeah, one translates to the other, I think I was stuck in pointer arithmetics at that point of the code

course6/string_ariel/main.cpp
```diff
@@ -0,0 +1,40 @@
+#include "str.h"
+
+int main()
+{
+    //Constructor
+    Str s1 = "foo";
+    std::cout << s1 << std::endl;
```
maciekgajewski: It's a valid interactive test, but in the end you want it to be automated. 
You can write a simple automated test using ```assert```, or do it properly with Google Test or Boost.Test

course6/string_ariel/str.h
```diff
@@ -0,0 +1,193 @@
...
+        {
+            *end++ = *s++;
+        }
+
+        mid = end;
+    }
+
+    Str(const Str& str)
+    {
+        int size = str.Size();
```
maciekgajewski: The content of this method is almost identical to the method above. There could be one function doing this.

course6/string_ariel/str.h
```diff
@@ -0,0 +1,193 @@
+#pragma once
+
+#include <iostream>
+#include <cstring>
+
+class Str
```
maciekgajewski: Wrap it into a namespace. Get into a habit of using namespaces. All the production code on the planet is inside of namespace, so it's good to get some experience with it as well.

course6/string_ariel/str.h
```diff
@@ -0,0 +1,193 @@
...
+        else
+        {
+            Str temp = *this + rh;
+            Swap(*this, temp);
+        }
+
+        return *this;
+    }
+
+    int Size() const
```
maciekgajewski: When implementing container, stick to the standard naming and interface. This way you can write generic code that can work on different containers, including yours.
Specifically, you should write your unit test as a template method, and it should pass for both your string and ```std::string```.

course6/string_ariel/str.h
```diff
@@ -0,0 +1,193 @@
...
+    {
+        char* index = start;
+        for (int i = 0; i < Size(); i++)
+        {
+            *index++ = str[i];
+        }
+    }
+};
+
+std::ostream &operator<<(std::ostream &os, const Str& str) {
```
maciekgajewski: This should be inline.
Or defined in .cc file. I think some of the bigger methods in this file should go to .cc.

course6/string_ariel/str.h
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    friend void Swap(Str& str1, Str& str2)
+    {
+        std::swap(str1.start, str2.start);
+        std::swap(str1.mid, str2.mid);
+        std::swap(str1.end, str2.end);
+    }
+
+    void Dispose()
```
maciekgajewski: Why separate method for that?

course6/string_ariel/str.h
```diff
@@ -0,0 +1,193 @@
...
+    {
+        delete[] start;
+    }
+
+    void Copy(const char* s)
```
maciekgajewski: It's called ```std::strcpy```, and you should use it in your constructors.
You wrote the same loop several times!

course6/string_ariel/str.h
maciekgajewski: Indeed. But containers usually have ```swap``` method as well.

course6/string_ariel/str.h
```diff
@@ -0,0 +1,193 @@
...
+            Str temp(s);
+            Swap(*this, temp);
+        }
+        else
+        {
+            mid = start + size;
+            Copy(s);
+        }
+
+        return *this;
```
maciekgajewski: again: two almost identical methods one after another

course6/string_ariel/str.h
```diff
@@ -0,0 +1,193 @@
...
+            }
+        }
+        else
+        {
+            Str temp = *this + rh;
+            Swap(*this, temp);
+        }
+
+        return *this;
+    }
```
maciekgajewski: and this is almost identical to the private concatenating constructor

maciekgajewski (changes requested):
You are missing one, very important method: c_str()

course6/string_ariel/str.h
a-moreno-a: I am not fully aware about how ADL works, but you meant that I should’ve put as public so it can be found using unqualified calls ?

course6/string_ariel/main.cpp
a-moreno-a: That’s true, did not have time to do a proper unit testing, and hesitated to add 3rd party libraries that cannot build for other people, at least I could’ve used assert tho

course6/string_ariel/str.h
a-moreno-a: Ok, will do on my next PR

course6/string_ariel/str.h
a-moreno-a: I am not completely sold, the concatenating constructor suits well inside the else statement (which is what the + operator does) but inside the if statement I am no really interested in allocating memory, that’s why I didnt use the constructor

course6/string_ariel/str.h
a-moreno-a: I tried to stay away as possible from container and some std stuff that we haven’t seen in the classes

a-moreno-a:
Agree with all the comments I didn’t reply

course6/string_ariel/str.h
maciekgajewski: No, ```swap``` should be free function, defined in the amer namespace as the object.
The good practice is to provide ```swap``` _method_, and implement swap _function_ in its terms:

```
class MyContainer
{
// ...
void swap(MyContainer& other) { /* ... */ }
};

void swap(MyContainer& a, MyContainer& b) { a.swap(b); }
```

course6/string_ariel/main.cpp
maciekgajewski: You should not hesitate to use Boost.
Also, if you not use any 3rd party lib, but simple ```assert``, then it will be simple and universal. Even easier to write and use than all these couts

course6/string_ariel/str.h
maciekgajewski: My point is: I see the same code twice (or even thrice). This shouldn't be. There is many ways of avoiding duplication.

course6/string_ariel/str.h
```diff
@@ -2,192 +2,100 @@
...
         {
-            *end++ = str2[i];
+            return Str(*this, rh);
         }
 
-        mid = end;
-    }
+        Str& operator += (const Str& rh)
+        {
+            Assign((*this + rh).start, this->size() + rh.size());
```
a-moreno-a: I know this is an extra allocatoin/deallocation but makes the code much more readable

maciekgajewski (approved):
Don't forget to implement c_str()!

course6/string_ariel/string_tests.cc
```diff
@@ -0,0 +1,76 @@
...
+
+using namespace mstr;
+
+BOOST_AUTO_TEST_SUITE (stringtest)
+
+BOOST_AUTO_TEST_CASE (constructor_tests)
+{
+  std::string s1 = "123";
+  Str t1 = "123";
+  BOOST_REQUIRE(s1.size() == t1.size());
```
maciekgajewski, line 16: Use ```BOOST_REQUIRE_EQUAL``` instead. This way the error message will contain the expected and got values.

course6/string_ariel/string_tests.cc
```diff
@@ -0,0 +1,76 @@
...
+#include <string>
+#include <cassert>
+#define BOOST_TEST_MODULE stringtest
+#include <boost/test/included/unit_test.hpp>
+
+using namespace mstr;
+
+BOOST_AUTO_TEST_SUITE (stringtest)
+
+BOOST_AUTO_TEST_CASE (constructor_tests)
```
maciekgajewski, line 12: Add test for default constructed string: make sure ```s.size() == 0```, ```std::strlen(s.c_str()) == 0```, ```s == ""``` etc

course6/string_ariel/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 cmake_minimum_required(VERSION 2.8)
 
 project(string_ariel)
+
+add_subdirectory(str)
```
maciekgajewski: You should add ``find_package(Boost REQUIRED ...)```, and then use variables set by it. Otherwise it is impossible to compile the code if Boost is installed outside of standard location

maciekgajewski: You need o set C++ standard to 14, otherwise the code won't compile on some compilers

maciekgajewski (changes requested):
Nice! Few small inefficiencies and (at least) one potentially crashing bug, but good!
Tests for empty string are missing, please, add them.

PR #187, opened 2017-10-03
Homework - Vector using template - Unittests

asmirnov911:


course6/toyapp_alexs/displaylib/vector.h
```diff
@@ -0,0 +1,29 @@
...
+
+	template<>
+	struct Vector<int, 2>
+	{
+		int x;
+		int y;
+	};
+
+	template<>
+	struct Vector<double, 3>
```
maciekgajewski: This is technically correct, but rather pointless.
The idea of templates is to allow user to use own types/sizes.

I was expecting something more like:
```
template<typename T, int S>
class Vector
{
    std::array<T, S> data;
public:
   // ...
};
```

course6/toyapp_alexs/displaylib/vector.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <array>
+
+namespace Display
+{
+	template<class T, std::size_t S>
```
maciekgajewski: Use ```typename``` rather than ```class```

course6/toyapp_alexs/displaylib/vector.h
```diff
@@ -0,0 +1,29 @@
...
+	struct Vector
+	{
+	};
+
+	template<>
+	struct Vector<int, 2>
+	{
+		int x;
+		int y;
+	};
```
maciekgajewski: To complete this taks, you should add some useful methods to the class: arithmetic operators, equality operators, streaming operator etc.

maciekgajewski (changes requested):
This is not what I expected

course6/toyapp_alexs/displaylib/matrix.h
```diff
@@ -1,37 +1,38 @@
 #pragma once
 
-#include "vector3d.h"
 #include <array>
 
 namespace Display {
 
-template<std::size_t A, std::size_t B>
+template<std::size_t A, std::size_t B, class T = double>
```
maciekgajewski, line 7: Prefer ```typename``` over ```class```. The latter is obsolete.

course6/toyapp_alexs/displaylib/matrix.h
```diff
@@ -1,37 +1,38 @@
...
 class Matrix
 {
 public:
 	Matrix() {}
-	Matrix(std::array<std::array<double, B>, A> data): mMatrix(data) {}
 
-	std::array<double, B>& operator [](int i)
+	template<class... E>
```
maciekgajewski, line 13: Template arguments don't have to be single letter, you know? :)

course6/toyapp_alexs/displaylib/matrix.h
```diff
@@ -1,37 +1,38 @@
...
+	Matrix(E... data): mMatrix {static_cast<T>(data)...} {}
+
+	std::array<T, B>& operator [](int i)
```
maciekgajewski, line 16: Does it return row or column? Is operator[] good API for returning row or column?

course6/toyapp_alexs/displaylib/vector.h
```diff
@@ -0,0 +1,79 @@
...
+	{}
+
+	Pixel& operator =(const Pixel& other)
+	{
+		this->x = other.x;
+		this->y = other.y;
+		return *this;
+	}
+
+	int& x;
```
maciekgajewski, line 46: Would be easier if you'd eschew the public fields, and provide getters ```x(); y()``` instead. 
Even better if you just acknowledge that you an always have API that describes the world precisely, and just have ```operator[]```.

course6/toyapp_alexs/displaylib/transformation3d.cc
```diff
@@ -98,12 +90,12 @@ Transformation3D Transformation3D::WithScaling(const Vector3D& vector) const
...
 Vector3D Transformation3D::Apply(const Vector3D& vector) const
 {
-	Matrix<4, 1> vMatrix {{
+	Matrix<4, 1> vMatrix {
 		vector.x,
 		vector.y,
 		vector.z,
 		1
-	}};
+	};
```
maciekgajewski, line 98: So having such a powerful Matrix, why don't you just implement vector like this:
```
template<typename T, int Size>
Vector = Matrix<Size, 1, T>;
```

Or if you need to add vector-specific operations:
```
template<typename T, int Size>
class Vector : public Matrix<Size, 1, T>
{
   // ...
};
```

maciekgajewski (approved):
Good!

PR #188, opened 2017-10-04
Homework: Templates

bodorkos-adam:


course6/toyapp-adam/math/vector.h
```diff
@@ -1,113 +1,172 @@
...
 
+#include <cassert>
+#include <array>
 #include <cmath>
+#include <limits>
 
 namespace Math {
 
-struct Vector2D
+template<typename Type, unsigned Dimensions, typename FloatType = float>
```
maciekgajewski: You don't need the ```FloatType``` to be the property of the entire class. You can use it as a template parameter on the methods that need it, so that you can use different floating-point types with the same Vector

course6/toyapp-adam/math/vector.h
```diff
@@ -1,113 +1,172 @@
...
+
+	operator Vector<Type, Dimensions-1>() const
+	{
+		Vector<Type, Dimensions-1> result;
+		for (unsigned i = 0; i < Dimensions-1; ++i)
+			result[i] = mComponents[i];
+		return result;
+	}
+
+	friend bool operator==(const Vector& lhs, const Vector& rhs)
```
maciekgajewski: This could be a member function. No need to use ```friend```

course6/toyapp-adam/math/vector.h
```diff
@@ -1,113 +1,172 @@
...
 	{
-		x += rhs.x;
-		y += rhs.y;
+		for (unsigned i = 0; i < Dimensions; ++i)
+			if (!ApproximatelyEquals(lhs[i], rhs[i]))
```
maciekgajewski: I think this is misleading. If your operator is called ```==```, your users will expect exact equality, floating-point quirks included.
You may provide additional method called ```ApproximatelyEquals``` for fuzzy comparison.

course6/toyapp-adam/math/vector.h
```diff
@@ -1,113 +1,172 @@
...
 	{
-		x -= rhs.x;
-		y -= rhs.y;
-		z -= rhs.z;
-		return *this;
+		return Vector{*this} += rhs;
 	}
 
-	Vector3D& operator*=(float scalar)
+	Vector operator*(FloatType scalar) const
```
maciekgajewski: Consider this simple form:
```
template<typename FloatType>
Vector operator*(FloatType scalar) const
{
   // ...
}
```
or, you you are feeling adventurous, you can either raise error if non-floating point type is used, lie that:
```
#include <type_traits>

template<typename FloatType>
Vector operator*(FloatType scalar) const
{
   static_assert(std::is_floating_point<FloatType>::value, "scalar must be float point value");
   // ...
}
```

or you can disable the function entirely for non-floating point types:

```
#include <type_traits>

template<typename FloatType, typename = std::enable_if_t<std::is_floating_point<FloatType>::value>>
Vector operator*(FloatType scalar) const
{
   // ...
}
```
The above is unreadable, you can make it easier by defining a helper:

```
#include <type_traits>
template<typename T>
using IsFloatingPointT = std::enable_if_t<std::is_floating_point<T>::value>;
```
then it will be easy to use, and readable:
```
template<typename FloatType, typename = IsFloatingPointT<FloatType>>
Vector operator*(FloatType scalar) const
{
   // ...
}
```

course6/toyapp-adam/math/vector.h
```diff
@@ -1,113 +1,172 @@
...
 
-	Vector3D Cross(const Vector3D& rhs) const
+	friend std::ostream& operator<<(std::ostream& stream, const Vector& vector)
 	{
-		return {
-			y * rhs.z - z * rhs.y,
-			z * rhs.x - x * rhs.z,
-			x * rhs.y - y * rhs.x
-		};
+		stream << vector[0] << ',' << vector[1];
```
maciekgajewski: Bug! won't work for vector with size != 2

course6/toyapp-adam/math/vector.h
```diff
@@ -1,113 +1,172 @@
...
+	{
+		using std::abs;
+		return abs(a - b) <= (abs(a) < abs(b) ? abs(b) : abs(a)) * std::numeric_limits<Type>::epsilon();
+	};
 };
 
-struct Vector4D
+
+template<typename Type>
+inline Vector<Type, 3> Cross(const Vector<Type, 3>& lhs, const Vector<Type, 3>& rhs)
```
maciekgajewski, line 169: nice!

maciekgajewski (approved):
Good!

course6/toyapp-adam/math/vector.h
bodorkos-adam: oops, oversight!

course6/toyapp-adam/math/vector.h
bodorkos-adam: I get compilation errors when I do that: `C:/lib/boost_1_65_1/boost/test/tools/old/impl.hpp:107:17: error: no match for 'operator==' (operand types are 'const Math::Vector<float, 2>' and 'const Math::Vector<float, 2>')
     return left == right;`

course6/toyapp-adam/math/vector.h
bodorkos-adam: You're right, this actually crossed my mind after I created the PR

course6/toyapp-adam/math/vector.h
maciekgajewski: was it a const method?

bodorkos-adam:
I don't know what the conflict is with `matrix.cc`

course6/toyapp-adam/math/vector.h
bodorkos-adam: I don't really agree. Although, I can't explain why. I might have to refresh my knowledge a bit about floats :) But I was thinking, because of the imprecision of floats, people wouldn't want to compare with bit-to-bit equality check. Maybe we could have a short explanation, discussion, refresher next time?

bodorkos-adam:
Updated this PR with the latest smart pointer homework (I had to modify `vector_tests.cc`)

course6/toyapp-adam/math/vector.h
```diff
@@ -116,6 +122,7 @@ class Vector
 		return Vector{*this} -= rhs;
 	}
 
+	template<typename FloatType = float, typename = IsFloatingPoint<FloatType>>
```
maciekgajewski: What's the point of making this a template?

course6/toyapp-adam/math/vector.h
```diff
@@ -69,6 +73,7 @@ class Vector
 		return *this;
 	}
 
+	template<typename FloatType = float, typename = IsFloatingPoint<FloatType>>
```
maciekgajewski: There is little point in giving the template param a default value. In most use-cases it will be deduced from call parameters.

course6/toyapp-adam/tests/util/unique_ptr_tests.cc
```diff
@@ -0,0 +1,89 @@
+#include "util/unique_ptr.h"
+#include <boost/test/unit_test.hpp>
+
+using namespace Util;
+
+namespace {
+	struct TestState
+	{
+		bool constructed = false;
+		bool destructed = false;
```
maciekgajewski, line 10: I think 'destroyed' would be the correct English term

maciekgajewski (approved):
Very good!

bodorkos-adam:
Thanks for checking! I looked at my code again now and I'm wondering if it works fine in case I use `Make` without any arguments. I'm not sure if it will pick the default constructor (nullptr) or the private, variadic one... but can't test right now.

course6/toyapp-adam/math/vector.h
bodorkos-adam: It uses `Length()` internally, which is a template.

PR #189, opened 2017-10-04
Homework week7

vivikelapoutre:
The vector template implementation works, though I think it should be done nicer (but still struggling how to pass arrays around). I can't get the unit tests to be build properly, unfortunately :(. Tried many things in how to write the cmake files, but nothing solved it.

### DIFF
--- a/course6/toyapp-vivike/displaylib/objects.h
+++ b/course6/toyapp-vivike/displaylib/objects.h
@@ -73,56 +73,19 @@ struct Vector
 	}
 };
 
-
-
-struct Vector2D
-{
-	int x;
-	int y;
-
-	void move(int xmove, int ymove);
-	
-	Vector2D operator +(Vector2D b) { return { x + b.x, y + b.y}; }
-	Vector2D operator +(int b) { return { x + b, y + b}; }	
-
-	Vector2D operator *(int b) { return { x * b, y * b}; }
-
-	Vector2D operator -(Vector2D r) { return { x - r.x, y - r.y}; }
-	Vector2D operator -(int r) { return { x - r, y - r}; }
-
-	Vector2D operator /(int r) { return { x / r, y / r}; }
-};
-
-struct Vector3D
-{
-	int x;
-	int y;
-	int z;
-
-	Vector3D operator +(Vector3D b) { return { x + b.x, y + b.y, z + b.z}; }
-	Vector3D operator +(int b) { return { x + b, y + b, z + b}; }	
-
-	Vector3D operator *(int b) { return { x * b, y * b, z * b}; }
-
-	Vector3D operator -(Vector3D r) { return { x - r.x, y - r.y, z - r.z}; }
-	Vector3D operator -(int r) { return { x - r, y - r, z - r}; }
-
-	Vector3D operator /(int r) { return { x / r, y / r, z / r}; }
-};
-
 struct Camera
 {
-	Vector2D project2D(Vector3D original);
+	Vector<int, 2> project2D(Vector<int, 3>& original);
 };
 
 struct Rectangle
 {
-	Vector3D corner1;
-	Vector3D corner2;
-	Vector3D corner3;
-	Vector3D corner4;
+	Vector<int, 3> corner1;
+	Vector<int, 3> corner2;
+	Vector<int, 3> corner3;
+	Vector<int, 3> corner4;
 
-	Rectangle move(Vector3D);
+	Rectangle move(Vector<int, 3>&);
 };
 
 }
--- a/course6/toyapp-vivike/displaylib/objects.h
+++ b/course6/toyapp-vivike/displaylib/objects.h
@@ -2,6 +2,7 @@
 
 #include <cstdint>
 #include <array>
+#include <iostream>
 
 namespace Display {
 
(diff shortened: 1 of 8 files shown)


course6/toyapp-vivike/displaylib/objects.h
```diff
@@ -14,44 +15,70 @@ struct Color
...
-	Vector3D operator +(Vector3D b) { return { x + b.x, y + b.y, z + b.z}; }
-	Vector3D operator +(int b) { return { x + b, y + b, z + b}; }	
-
-	Vector3D operator *(int b) { return { x * b, y * b, z * b}; }
-
-	Vector3D operator -(Vector3D r) { return { x - r.x, y - r.y, z - r.z}; }
-	Vector3D operator -(int r) { return { x - r, y - r, z - r}; }
-
-	Vector3D operator /(int r) { return { x / r, y / r, z / r}; }
+	int vec[length];
```
maciekgajewski: Using ```std::array``` would make things much easier

course6/toyapp-vivike/displaylib/objects.h
```diff
@@ -14,44 +15,70 @@ struct Color
...
+
+	Vector<T, length> operator +(Vector& b)
```
maciekgajewski: Inside of the class' definition, you don't need to provide template arguments for the class. ```Vector``` will mean ```Vector<T, length>```

course6/toyapp-vivike/displaylib/objects.h
```diff
@@ -14,44 +15,70 @@ struct Color
...
+	{ 
+		Vector<T, length> newVec;
+		for(int i=0; i<length; i++)
+		{
+			newVec.vec[i] = this->vec[i] / r;
+		}
+		return newVec;
+	}	
+
+	void print()
```
maciekgajewski: Use ```std::ostream& operator<<(std::ostream&, const Vector&)``` instead. It is much more flexible

course6/toyapp-vivike/displaylib/objects.h
```diff
@@ -14,44 +15,70 @@ struct Color
...
+		{
+			std::cout << v << std::endl;
+		}
+		
+
+	}
 };
 
+using Pixel = Vector<int, 2>;
+using Vector3D = Vector<float, 3>;
```
maciekgajewski: Exactly!

course6/toyapp-vivike/displaylib/objects.h
maciekgajewski: This should be private. You don't want to expose your implementation detail.

course6/toyapp-vivike/displaylib/surface.cc
```diff
@@ -24,57 +26,62 @@ void Surface::Clear(Color backgroundColor)
...
+	std::uint32_t* pixelAddr = static_cast<std::uint32_t*>(mSurface->pixels) + (pixel.vec[1]*mSurface->pitch/4 + pixel.vec[0]);
 	*pixelAddr = pixelValue;
 }
 
-void Surface::DrawLine(Vector2D p1, Vector2D p2, Color color)
+void Surface::DrawLine(Pixel& p1, Pixel& p2, Color color)
 {
 	// Bresenham's line algorithm
-	const bool steep = (fabs(p2.y - p1.y) > fabs(p2.x - p1.x));
+	int v1x = p1.vec[0];
```
maciekgajewski: This is awkward API. It would be better to implement ```operator[]``` in your vector, so that uses can use it like that:
```
int v1x = p1[0];
```

maciekgajewski (changes requested):
Mostly good!
Fix the compilation issues please.

course6/toyapp-vivike/displaylib/objects.h
vivikelapoutre: I'm working on it, but just to be sure, if I want to make vec private, I need a constructor to still be able to handle something like ```Vector3D myVector{1, 1, 1}```?

course6/toyapp-vivike/tests/test_objects.cc
```diff
@@ -10,21 +10,57 @@ BOOST_AUTO_TEST_SUITE (vector_tests)
 
 BOOST_AUTO_TEST_CASE (constructor_tests)
 {
-  Vector<int, 3> myvec{1, 2, 3};
-  BOOST_REQUIRE(myvec.vec[0] == 1);
-  BOOST_REQUIRE(myvec.vec[1] == 2);
-  BOOST_REQUIRE(myvec.vec[2] == 3);
+  Vector<int, 3> myvec{3, 9, 6};
+  BOOST_REQUIRE(myvec[0] == 3);
```
maciekgajewski: Use ```BOOST_REQUIRE_EQUAL``` instead, it will produce nicer error message.

course6/toyapp-vivike/displaylib/objects.h
```diff
@@ -14,44 +15,79 @@ struct Color
...
+template<typename T, int length>
+class Vector
 {
-	int x;
-	int y;
-	int z;
+public:
+	std::array<T, length> vec;
+
+	//Vector(std::array<T, length> newvec): vec(newvec) {};
```
maciekgajewski, line 25: This would work, you'd have to use it like that:
```
Vector b({1, 2, 3});
```

Alternatively, you can make the constructor a variadic template. Look at the code of other participants.

maciekgajewski (commented):
Good

PR #190, opened 2017-10-04
Mikhails/homework template

Mikhail-SHA-256:
This is just template implementation part of homework

### DIFF
--- a/course6/toyapp_mikhail/displaylib/vector2d.h
+++ b/course6/toyapp_mikhail/displaylib/vector2d.h
@@ -1,26 +1,7 @@
 #pragma once
 
+#include "vector_common.h"
+
 namespace Display {
-
-	class Vector2D
-	{
-	public:
-		double x, y;
-
-		Vector2D operator + (const Vector2D& v)
-		{
-			auto r = Vector2D(*this);
-			r.x += v.x;
-			r.y += v.y;
-			return r;
-		}
-
-		Vector2D operator * (double scale)
-		{
-			auto r = Vector2D(*this);
-			r.x *= scale;
-			r.y *= scale;
-			return r;
-		}
-	};
+	using Vector2D = VectorCommon<double, 2>;
 }
--- a/course6/toyapp_mikhail/displaylib/vector2d.h
+++ b/course6/toyapp_mikhail/displaylib/vector2d.h
@@ -1,9 +1,0 @@
-#pragma once
-
-#include "vector_common.h"
-
-namespace Display {
-
-using Vector2D = VectorCommon<double, 2>;
-
-}
--- a/course6/toyapp_mikhail/displaylib/vector2d.h
+++ b/course6/toyapp_mikhail/displaylib/vector2d.h
@@ -3,5 +3,7 @@
 #include "vector_common.h"
 
 namespace Display {
-	using Vector2D = VectorCommon<double, 2>;
+
+using Vector2D = VectorCommon<double, 2>;
+
 }
--- a/course6/toyapp_mikhail/displaylib/vector_common.h
+++ b/course6/toyapp_mikhail/displaylib/vector_common.h
@@ -4,7 +4,7 @@
 
 namespace Display {
 
-namespace Detail{
+namespace Detail {
 template <class TValue, size_t DIMENTIONS, class TDecoratedType>
 class VectorBase
 {
@@ -36,7 +36,7 @@ public:
 
 // 2 dimentional template vector. Implementetion needed only to provide x(), y() interface.
 template <class TValue>
-class Vector2 : public Detail::VectorBase<TValue, 2, Vector2DimDecorator<TValue>>
+class Vector2 : public Detail::VectorBase<TValue, 2, Vector2<TValue>>
 {
 private:
 	using TBase = Detail::VectorBase<TValue, 2, Vector2<TValue>>;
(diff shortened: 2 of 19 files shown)


course6/toyapp_mikhail/displaylib/vector2d.h
```diff
@@ -1,26 +1,9 @@
...
+namespace Display {
 
-		Vector2D operator + (const Vector2D& v)
-		{
-			auto r = Vector2D(*this);
-			r.x += v.x;
-			r.y += v.y;
-			return r;
-		}
+using Vector2D = VectorCommon<double, 2>;
```
maciekgajewski: I would say it's statistical overkill to have a separate source file for one line

course6/toyapp_mikhail/displaylib/vector_common.h
```diff
@@ -0,0 +1,57 @@
+#pragma once
+
+namespace Display{
+
+template <class TValue, int TDimentions>
+class VectorCommon
+{
+	// No constructor for vector without specification
```
maciekgajewski: Simply provide no definition for primary template at all, then only specializations will be allowed. 
But in this case I was expecting an actual primary template that works with any size

course6/toyapp_mikhail/displaylib/vector_common.h
```diff
@@ -0,0 +1,57 @@
...
+	VectorCommon() = delete;
+};
+
+// Specification for 2 dimentional vector
+template <class TValue>
```
maciekgajewski: no no no, this is not what I had in mind. The goal of this template should be to use vector of any size/type

maciekgajewski (changes requested):
This is not what I expected. See other's solutions

course6/toyapp_mikhail/displaylib/vector2d.h
Mikhail-SHA-256: Fixed. Left just one file with template definitions

course6/toyapp_mikhail/displaylib/vector_common.h
Mikhail-SHA-256: Changed.
But I still want to use nice decorations with x,y and x,y,z to not change client code.
Using v.x, v.y in client code is cleaner than v[0], v[1].
For decorations I had to use inheritance. is it ok?

maciekgajewski: Forget about ```v.x```, ```v.y```. C++ doesn't have the support for properties.
The canonical way is to use accessor methods: getters and setters.

You may implement ```v.x()```, ```v.y()``` in an inherited class, and they can be an alternative api to ```v[0]```.

You will have to change your client code one way or the other. Consider this a punishment for using public fields :)

Mikhail-SHA-256: Ok

Mikhail-SHA-256:
Add more fixes based on feedback

Separate commit for reformatting

course6/toyapp_mikhail/displaylib/vector_common.h
```diff
@@ -0,0 +1,100 @@
...
+
+#include <array>
+
+namespace Display {
+
+template <class TValue, size_t TDimentions, class TDecoratedType = TValue>
+class VectorCommon
+{
+private:
+	using TSelf = VectorCommon<TValue, TDimentions>;
```
maciekgajewski: You can just use ```VectorCommon``` inside of the class' definition.

course6/toyapp_mikhail/displaylib/vector_common.h
```diff
@@ -0,0 +1,100 @@
+#pragma once
+
+#include <array>
+
+namespace Display {
+
+template <class TValue, size_t TDimentions, class TDecoratedType = TValue>
```
maciekgajewski: the "T" in "TDimensions" doesn't styand for "type"? right?
The popular convention is to use ALL_CAPS for value template arguments. Just like for constants; they are compile-time constants after all.

maciekgajewski: Why is "TDecoratedType"? It's not imemdiately obvious for mew why would you need that in a vector, which means it needs eithjer better name, or at least explanation in a comment.

maciekgajewski: Ah! Decorated type is the implementation! You are doing CRTP! Then using TValue as default makes no sense, and TSelf is useless.

course6/toyapp_mikhail/displaylib/vector_common.h
```diff
@@ -0,0 +1,100 @@
...
+		(*this)[1] = yy;
+		(*this)[2] = zz;
+	}
+
+	Vector3DimDecorator(const TBase& original) :
+		Vector3DimDecorator(original[0], original[1], original[2])
+	{}
+};
+
+using Vector2D = Vector2DimDecorator<double>;
```
maciekgajewski: Decorator/Decorated is not a canonical naming convention.
```Vector2Dim``` would be a better name, and ```VectorCommon``` should rather be ```VectorBase```

maciekgajewski (changes requested):
Good use of CRTP!
But the naming and some details make it confusing.

course6/toyapp_mikhail/displaylib/vector_common.h
Mikhail-SHA-256: namespace Display::Detail

maciekgajewski (approved):
Good!

PR #191, opened 2017-10-04
Implemented the vector template, just the basic operations

a-moreno-a:
TODO: integrate in graphics api and unit test

### DIFF
--- a/course6/toyapp_ariel/displaylib/vector.h
+++ b/course6/toyapp_ariel/displaylib/vector.h
@@ -10,6 +10,20 @@ class Vector
 {
 public:
     Vector(const std::array<T, size>& v) :  values(v) {}
+    Vector(int n)
+    {
+        std::for_each(values.begin(), values.end(), [n](T& current){current = n;});
+    }
+
+    T operator[](int i) const
+    {
+        return values[i];
+    }
+
+    T& operator[](int i)
+    {
+        return values[i];
+    }
 
     T Length() const
     {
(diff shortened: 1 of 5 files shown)


course6/toyapp_ariel/displaylib/vector.h
```diff
@@ -0,0 +1,63 @@
...
+
+    T Length() const
+    {
+        return std::sqrt(Dot(*this));
+    }
+
+    T Dot(const Vector& v) const
+    {
+        auto temp = TransformVectors<std::multiplies<T>>(v);
+        return std::accumulate(temp.begin(), temp.end(), 0, std::plus<T>());
```
maciekgajewski: Interesting, but old-school. Lambdas would be better than functors,

course6/toyapp_ariel/displaylib/vector.h
```diff
@@ -0,0 +1,63 @@
...
+    Vector operator + (const Vector& v) const
+    {
+        return TransformVectors<std::plus<T>>(v);
+    }
+
+private:
+    std::array<T, size> values;
+
+    template<typename F>
+    Vector TransformVectors(const Vector& v) const
```
maciekgajewski: Take F as parameter, so that it can be deduced. Then use lambdas. Will be more convenient and flexible

maciekgajewski (changes requested):
Interesting use of algorithms, but I'd like to see unit tests for the vector. for different sizes and types

maciekgajewski (approved):
Nice!

PR #193, opened 2017-10-16
Homework - OBJ mesh rendering

bodorkos-adam:
I have a huge issue:
* for some weird reason, Debug build works OK (but rendering is very slow and laggy), but Release build segfaults on this line: `std::ifstream file(filePath);`. I couldn't figure out what the problem is. At some point earlier though I did manage to load files from a Release build (don't know what triggered the change in behavior) and the rendering wasn't lagging at all (7% CPU usage on 3.5 Ghz core).

Other:
* only supports a subset of OBJ files at this point, having normal / texcoord indices causes an empty mesh
* triangle rendering is still imperfect, there is overdraw and holes
* I draw edges to cover that, but there's Z-fighting with those. I could draw all triangles first and then the edges to fix though


course6/toyapp-adam/displaylib/mesh_util.cc
```diff
@@ -0,0 +1,65 @@
...
+static unsigned GetFaceIndex(unsigned index, unsigned vertexCount)
+{
+	return index > 0 ? index - 1 : vertexCount + index;
+}
+
+Mesh CreateMeshFromObjFile(const std::string& filePath)
+{
+	Mesh mesh;
+
+	std::ifstream file(filePath);
```
maciekgajewski, line 18: So does it segfaults or throws an exception?

course6/toyapp-adam/displaylib/surface_2d.cc
```diff
@@ -3,12 +3,14 @@
...
 #include <cassert>
 #include <cmath>
+#include <limits>
 
 namespace Display {
 
 Surface2D::Surface2D(Pixel dimensions)
 :	mHalfDimensions{dimensions * 0.5f}
 {
+	mDepthBuffer = std::make_unique<float[]>(dimensions[0] * dimensions[1]);
```
maciekgajewski, line 13: Why don't you use ```std::vector``` here?

course6/toyapp-adam/displaylib/surface_2d.cc
```diff
@@ -22,12 +24,35 @@ void Surface2D::Clear(Color color)
...
+		{
+			Pixel pixel{x, y};
+			SetPixel(pixel, color);
+			SetDepth(pixel, std::numeric_limits<float>::max());
+		}
+}
+
+void Surface2D::SetDepth(Pixel position, float depth)
+{
+	mDepthBuffer[position[1] * mSurface->w + position[0]] = depth;
```
maciekgajewski, line 36: It looks to me like the DepthBuffer deserves to be a class by itself. You already have few methods for it, that are implemented in another object.

course6/toyapp-adam/scene.cc
```diff
@@ -1,29 +1,30 @@
...
-	mTestCube5{{0.0f, 1.125f, 2.5f}, 0.125f, Display::YELLOW, Display::BLACK}
+	mSurface{surface}
 {}
 
-void Scene::Update(double totalElapsedSeconds)
+void Scene::LoadMesh(const std::string& filePath)
 {
-	using Math::Vector3D;
+	mMesh = Display::CreateMeshFromObjFile(filePath);
+	mMesh.NormalizeScale();
```
maciekgajewski, line 17: I don't like this "normalise scale" thing. It mutates the mesh to fit your viewport. It should be the other way around.
Maybe the mesh should have a method that returns AABB, and then you can adjust your camera so that all of the mesh is visible?

course6/toyapp-adam/displaylib/mesh_util.cc
```diff
@@ -0,0 +1,65 @@
...
+	std::vector<Math::Vector3D> vertices;
+
+	std::string line;
+	while (std::getline(file, line))
+	{
+		if (line.length() < 2)
+			continue;
+
+		std::stringstream stream(line);
+		if (line[0] == 'v')
```
maciekgajewski, line 31: why don't you just read ```char``` ot ```std::string``` from the stream? IT would be more robust, and you wouldn't have to call ignore()

maciekgajewski (approved):
nice!

PR #194, opened 2017-10-18
Homework week3 transformations

vivikelapoutre:
Finally gaining a bit on late homework.
I tried to make the matrix class print (by the bit of commented out code), but if I uncomment it, it says it is already defined, but I don't see where that could be.


course6/toyapp-vivike/displaylib/vector.h
```diff
@@ -0,0 +1,128 @@
...
+		result += this->vec[i] * vecb[i];
+	}
+	return result;
+}
+
+using Pixel = Vector<int, 2>;
+using MatrixElement = Vector<int, 2>;
+using Vector3D = Vector<float, 3>;
+
+/* Don't think I want to go through the effort of implementing this, since it only works in 3D.
```
maciekgajewski, line 117: Maybe this should be a stand-alone function that works for 3-element vectors only?

course6/toyapp-vivike/displaylib/matrix.h
```diff
@@ -0,0 +1,54 @@
...
+
+	static Matrix44 Zero();
+	static Matrix44 Eye();
+
+	static Matrix44 Scale(float scaling);
+	static Matrix44 Scale(Vector3D scaling);
+	static Matrix44 Translate(Vector3D translation);
+};
+
+/*std::ostream& operator <<(std::ostream& stream, Matrix44 mat)
```
maciekgajewski, line 41: of course you get "multiple definition" _linker_ errors here

maciekgajewski (approved):
Quite good. Too bad you gave up on the dot and operator<<

PR #197, opened 2017-10-29
Meeting 11 homework

mtux:
Please ignore the old stuff outside of meeting11 folder.


course6/meeting11/mehrdad-function/function.h
```diff
@@ -0,0 +1,68 @@
...
+	};
+
+	template<typename F>
+	struct Callable : BaseCallable
+	{
+		Callable(const F& f) : mF(f) {}
+
+		R Invoke(Args... args) override
+		{
+			return mF(args...);
```
maciekgajewski, line 59: You should use forwarding references and ```std::forward``` here. This exposes a flaw in your testing: you should test your function with different parameters - including rvalue refs, references, non-copyable objects and other exotics.

course6/meeting11/mehrdad-function/function.h
```diff
@@ -0,0 +1,68 @@
...
+	template<typename F>
+	Function(F f)
+	{
+		mCallable = std::make_unique<Callable<F>>(f);
+	}
+
+	R operator() (Args... args)
+	{
+		if (!mCallable)
+			throw std::runtime_error("Calling empty function");
```
maciekgajewski, line 33: this should be ```std::bad_function```

maciekgajewski (approved):
Do you have to include catch? Can;t you use it as an external? Or just require it to be installed to compile the test?

PR #199, opened 2017-11-06
Add parsing of shape (obj) file and render all triangles on surface

Mikhail-SHA-256:
Catch up with cow rendering


course6/toyapp_mikhail/displaylib/shape.cc
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Display {
+
+Shape Shape::LoadFromFile(const std::string& filePath)
+{
+	Shape shape;
+
+	std::ifstream file(filePath);
+	if (file.fail())
+		return shape;
```
maciekgajewski, line 17: Why don't you throw an exception?

maciekgajewski (changes requested):
The code looks good, but it doesn't compile neither with gcc, nor with clang

PR #203, opened 2018-01-22
Add ls as first homework implementation

patrickkostjens:


course7/homeworks/patrickkostjens/ls/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+cmake_minimum_required(VERSION 3.10)
```
maciekgajewski, line 1: You are not using any CMake features introduced in 3.10. It's better to use lower number here, the lowest possible.
3.0 is a reasonable value.
Otherwise, people with slightly older version won't be able to use it!

course7/homeworks/patrickkostjens/ls/ls.cc
```diff
@@ -0,0 +1,13 @@
+#include <iostream>
+#include <boost/filesystem.hpp>
```
maciekgajewski, line 2: It's a good practice to order header from the most local to the most generic.
So:
- this file's header
- this project's header
- your libs
- 3rd party libs
- boost
- std

course7/homeworks/patrickkostjens/ls/ls.cc
```diff
@@ -0,0 +1,13 @@
...
+
+int main(int argc, char** argv)
+{
+	boost::filesystem::path current_directory(argc >= 2 ? argv[1] : ".");
+
+	boost::filesystem::directory_iterator end;
+	for (boost::filesystem::directory_iterator itr(current_directory); itr != end; ++itr)
```
maciekgajewski, line 9: This is too old school. Is this because you are using very old version of Boost?
Please, have a look at Waleed's implementation to see how to use range-based-for for this.

course7/homeworks/patrickkostjens/ls/ls.cc
```diff
@@ -0,0 +1,13 @@
...
+#include <boost/filesystem.hpp>
+
+int main(int argc, char** argv)
+{
+	boost::filesystem::path current_directory(argc >= 2 ? argv[1] : ".");
+
+	boost::filesystem::directory_iterator end;
+	for (boost::filesystem::directory_iterator itr(current_directory); itr != end; ++itr) 
+	{
+		std::cout << itr->path().filename().string() << std::endl;
```
maciekgajewski, line 11: is the call to ```string()``` necessary?

maciekgajewski (approved):
Correct

course7/homeworks/patrickkostjens/ls/ls.cc
patrickkostjens: Not strictly necessary, but if it's not added, all the elements will be printed with double quotes around them.

PR #204, opened 2018-01-22
Simple "ls"

waleedlotfy:


course7/homeworks/waleed/ls/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.9)
```
maciekgajewski: If you're not using any feature that was introduced in CMake 3.9, then it's better yo use lower version (3.0 for instance), so that people who happen to have Cmake 3.8 also can build the project.

course7/homeworks/waleed/ls/main.cc
```diff
@@ -0,0 +1,16 @@
+#include <boost/filesystem.hpp>
+
+#include <iostream>
+
+int main(int argc, char** argv)
+{
+	boost::filesystem::path CurrPath = boost::filesystem::current_path();
```
maciekgajewski: Inconsistent naming. This variable is starts with CAPS, while argc/argv start with lower caps.

maciekgajewski (changes requested):
Almost perfect :)

PR #205, opened 2018-01-23
hw1: simple ls

anton-kultsov:


course7/homeworks/anton/ls/homework01.cc
```diff
@@ -0,0 +1,14 @@
+#include <boost/filesystem.hpp>
+#include <iostream>
+
+int main(int argc, char *argv[])
+{
+    boost::filesystem::path path(".");
+
+    for(auto& dir_entry: boost::filesystem::directory_iterator(path))
+    {
+        std::cout << dir_entry.path().filename().string() << std::endl;
```
maciekgajewski, line 10: I don't think the call to ```string()``` is required

maciekgajewski (approved):
This is it!

course7/homeworks/anton/ls/homework01.cc
anton-kultsov: it is not required, but output differs. Without the call to `string`, filenames are double-quoted.

PR #206, opened 2018-01-24
Simple ls

MartSlot:
No fancy stuff as I was sick and didn't have the time

### DIFF
--- a/course7/homeworks/mart/ls/CMakeLists.txt
+++ b/course7/homeworks/mart/ls/CMakeLists.txt
@@ -1,0 +1,3 @@
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -std=c++14")
+
+add_executable(ls ls.cc)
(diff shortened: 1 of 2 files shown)


maciekgajewski:
"no fancy stuff" is exactly what I asked for

course7/homeworks/mart/ls/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.5)
```
maciekgajewski, line 1: Too high. Use 3.0

course7/homeworks/mart/ls/ls.cc
```diff
@@ -0,0 +1,13 @@
+#include <iostream>
+#include <boost/filesystem.hpp>
```
maciekgajewski, line 2: Order includes from the most local one to the most general one. So Boost should be before std

course7/homeworks/mart/ls/ls.cc
```diff
@@ -0,0 +1,13 @@
...
+
+int main(int argc, char** argv)
+{
+	auto current_path = boost::filesystem::current_path();
+
+	for (auto& entry : boost::filesystem::directory_iterator(current_path))
+	{
+		auto filename = entry.path().filename();
+		std::cout << filename.string() << std::endl;
```
maciekgajewski, line 11: Is the call to ```string``` required? doesn't the ```path``` class has an ostream operator?

maciekgajewski (approved):
Good. As simple as requested.

PR #207, opened 2018-01-24
Homework 1

Pvanhesteren:


course7/homeworks/patrickh/ls/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.0.0)
+set(CMAKE_CXX_STANDARD 14)
+project(ls)
+
+set(Boost_USE_STATIC_LIBS        ON)
+set(Boost_USE_MULTITHREADED      ON)
+set(Boost_USE_STATIC_RUNTIME    OFF)
+find_package(Boost COMPONENTS filesystem REQUIRED)
+
+if(Boost_FOUND)
```
maciekgajewski, line 10: If you requested Boost as a REQUIRED in the ```find_package``` above, then there is no point checking if it was found.

course7/homeworks/patrickh/ls/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.0.0)
+set(CMAKE_CXX_STANDARD 14)
+project(ls)
+
+set(Boost_USE_STATIC_LIBS        ON)
+set(Boost_USE_MULTITHREADED      ON)
+set(Boost_USE_STATIC_RUNTIME    OFF)
```
maciekgajewski, line 7: I don't think this should be needed

course7/homeworks/patrickh/ls/ls.cc
```diff
@@ -0,0 +1,14 @@
+#include <iostream>
+#include <boost/filesystem.hpp>
+
+int main()
+{
+	std::string path (".");
+	if (boost::filesystem::is_directory(path))
```
maciekgajewski, line 7: What else could "." be?

course7/homeworks/patrickh/ls/ls.cc
```diff
@@ -0,0 +1,14 @@
...
+	{
+		for (boost::filesystem::directory_iterator iterator(path); iterator!=boost::filesystem::directory_iterator(); ++iterator)
```
maciekgajewski, line 9: It would be much easier to read if you'd use range-based-for here. Look at solutions submitted by others.

maciekgajewski (commented):
Bit too much, but correct

PR #209, opened 2018-01-25
Assignment1

nadinasovaiala:


course7/homeworks/nadinasovaiala/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.8)
```
maciekgajewski, line 1: This is too high. Someone with CMake 3.1 could easily use it, because the file doasn't xcontain any super-new Cmake features, but you prevent this.
Use 3.0 at the most

course7/homeworks/nadinasovaiala/assign1.cc
```diff
@@ -0,0 +1,12 @@
+#include <iostream>
+#include <boost/filesystem.hpp>
+
+using namespace boost::filesystem;
```
maciekgajewski, line 4: Don't do it.
If you think ```boost::filesystem``` is too long, create a shorter alias:
```
namespace bfs = boost::filesystem;
```
but don't drop the qualification altogether.

maciekgajewski (approved):
Good

PR #210, opened 2018-01-25
Kush

ksing:
Finally figured out how github works!

### DIFF
--- a/course7/homeworks/kushagra/CMakeLists.txt
+++ b/course7/homeworks/kushagra/CMakeLists.txt
@@ -1,11 +1,14 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.0)
 
 project(my_ls)
 
-find_package(Boost COMPONENTS filesystem REQUIRED)
+find_package(Boost 1.55 REQUIRED COMPONENTS filesystem system)
+
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fexceptions -Wextra -std=c++14")
-add_executable(my_ls my_ls.cpp)
+if(Boost_FOUND)
+	include_directories(${Boost_INCLUDE_DIRS})
+	add_executable(my_ls my_ls.cpp)
+	target_link_libraries(my_ls ${Boost_LIBRARIES})
+endif()
 
-target_link_libraries(my_ls PRIVATE Boost::filesystem)                                                           
-target_include_directories(my_ls PRIVATE Boost::boost)             
--- a/course7/homeworks/kushagra/CMakeLists.txt
+++ b/course7/homeworks/kushagra/CMakeLists.txt
@@ -1,0 +1,11 @@
+cmake_minimum_required(VERSION 2.8)
+
+project(my_ls)
+
+find_package(Boost COMPONENTS filesystem REQUIRED)
+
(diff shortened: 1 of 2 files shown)


course7/homeworks/kushagra/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.0)
+
+project(my_ls)
+
+find_package(Boost 1.55 REQUIRED COMPONENTS filesystem system)
+
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fexceptions -Wextra -std=c++14")
```
maciekgajewski, line 8: Don't set too many compiler-specific flags, and definitely never set -g. Leave it to the person building your project to decide whether it should be a release or debug build.

course7/homeworks/kushagra/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+if(Boost_FOUND)
```
maciekgajewski, line 9: Why the if if you mark Boost as REQUIRED in find_package above. If Boost is not found, configuration will fail altogether.

course7/homeworks/kushagra/my_ls.cpp
```diff
@@ -0,0 +1,31 @@
+#include <iostream>
+#include <boost/filesystem.hpp>
```
maciekgajewski, line 2: Order your includes from the most local to the most generic ones. So the STD includes should always be a the end.

course7/homeworks/kushagra/my_ls.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+int main(int argc, char** argv)
+{
+	boost::filesystem::path full_path(boost::filesystem::current_path());
+    if (argc > 1) {
+        boost::filesystem::path new_path(argv[1]);
+		full_path.swap(new_path) ;
```
maciekgajewski, line 9: Something is wrong with the indentation here.

maciekgajewski: no need to swap, you can just assign

course7/homeworks/kushagra/my_ls.cpp
```diff
@@ -0,0 +1,31 @@
...
+                    std::cout << "\t" << x.path() << std::endl;
+                } 
+            } 
+        }
+        else {
+            std::cout << full_path << std::endl;
+        }
+    }
+    catch (const boost::filesystem::filesystem_error& exception) {
+        std::cout << exception.what() << std::endl;
```
maciekgajewski, line 28: Don't catch it. Let it fly trough and kill the program with a meaningful coredump containing the stack trace to the place where the exception was thrown. AEs will thank you for that :)

course7/homeworks/kushagra/my_ls.cpp
```diff
@@ -0,0 +1,31 @@
...
+                std::cout << full_path << std::endl;
+            }
+            else {
+                for (boost::filesystem::directory_entry& x : boost::filesystem::directory_iterator(full_path)) {
+                    std::cout << "\t" << x.path() << std::endl;
+                } 
+            } 
+        }
+        else {
+            std::cout << full_path << std::endl;
```
maciekgajewski, line 24: This is some debug code, right? It should be removed from the final solution.

maciekgajewski (changes requested):
The solution is not complete. There is still some code that looks like debug helper in there.

PR #211, opened 2018-01-27
Homework 2: Extend ls with more options

patrickkostjens:


course7/homeworks/patrickkostjens/ls/ls.cc
```diff
@@ -1,13 +1,60 @@
 #include <boost/filesystem.hpp>
+#include <boost/program_options.hpp>
 #include <iostream>
 
+namespace fs = boost::filesystem;
+namespace po = boost::program_options;
+
+size_t get_size(fs::path path)
```
maciekgajewski: ```fs::path``` is a class  that's potentially expensive to copy. You should always try to pass such objects by const reference, to save the copying

course7/homeworks/patrickkostjens/ls/ls.cc
```diff
@@ -1,13 +1,60 @@
...
+
+	size_t size = 0;
+	for (const auto& it : fs::directory_iterator(path)) 
+		size += get_size(it);
+	return size;
+}
+
+void print_contents(fs::path path, bool extensive, bool tree, int level = 0)
+{
+	const auto& is_directory = fs::is_directory(path);
```
maciekgajewski: ```is_directory``` returns simple boolean, a 1-byte number. There is no point storing a reference to it.

course7/homeworks/patrickkostjens/ls/ls.cc
```diff
@@ -1,13 +1,60 @@
...
+
+	std::cout << std::string(level * 2, ' ') << path.filename().string();
+	if (is_directory) std::cout << "/";
+	if (extensive) std::cout << ": " << get_size(path);
+	std::cout << std::endl;
+
+	if (is_directory && tree)
+		for (const auto& it : fs::directory_iterator(path))
```
maciekgajewski: So here you descend recursively down the tree _again_. This means that you will call ```fs::file_size``` many times on the same file. This is an IO operation, potentially very slow.

Can you try re-structure the program so that you only visit every file once?

maciekgajewski (changes requested):
Good, but:
- you use const ref when it's not needed
- you don't use const ref when it's needed
- you revisit the same file many times, making the program very inefficient

patrickkostjens:
Every file is now visited at most twice, I'll try to find a neat way to visit every file only once.

course7/homeworks/patrickkostjens/ls/ls.cc
```diff
@@ -1,13 +1,62 @@
...
+
+struct node_result
+{
+	std::string output;
+	size_t size;
+};
+
+node_result* print_contents(const fs::path& path, bool extensive, bool tree, int level = 0)
+{
+	node_result* result = new node_result();
```
maciekgajewski: Don't! This is C#-ism. Just create the object on the stack and return it as if it was a normal value.
Please mote: you never delete the object, leaking resources!

maciekgajewski (changes requested):
Nice, but don't use 'new'! Never, unless you are working on some low-level library.

patrickkostjens:
Should be fixed. Haven't had time to compare performance of string and stringstream.

maciekgajewski (approved):

PR #212, opened 2018-01-28
Homework 2

waleedlotfy:


course7/homeworks/waleed/ls/main.cc
```diff
@@ -1,16 +1,128 @@
...
+namespace bfs = boost::filesystem;
+
+unsigned int fileSize(const bfs::path& path)
 {
-	boost::filesystem::path currPath = boost::filesystem::current_path();
-	boost::filesystem::directory_iterator currDir(currPath);
+	if (!bfs::exists(path))
+		throw std::runtime_error("File doesn't exist");
+
+	if (bfs::is_regular_file(path))
```
maciekgajewski, line 16: ```is_regular_file``` will return false also for files that are not directories, like device nodes or symlinks.

course7/homeworks/waleed/ls/main.cc
```diff
@@ -1,16 +1,128 @@
...
+	std::stringstream ss;
+	if (bfs::is_regular_file(path))
+	{
+		ss << treeTraversal(path, treeArg, longArg);
+	}
+	else
+	{
+		bfs::directory_iterator currDir(path);
+
+		for (const auto& it : currDir)
```
maciekgajewski, line 78: You have 3 loops like this. I wonder if the program can be written with only one loop.

course7/homeworks/waleed/ls/main.cc
```diff
@@ -1,16 +1,128 @@
...
 
 	for (const auto& it : currDir)
 	{
-		std::cout << it.path().filename() << std::endl;
+		result += fileSize(it.path());
+	}
+	return result;
+}
+
+std::string treeTraversal(const bfs::path& path, const bool& treeArg, const bool& longArg, unsigned int indentLevel = 0)
```
maciekgajewski, line 31: There is no point passing booleans as references. Boolean is a primitive type, one byte, cheap to copy. Cheaper, in fact, than 8 bytes of reference!

course7/homeworks/waleed/ls/main.cc
```diff
@@ -1,16 +1,128 @@
...
+{
+	if (!bfs::exists(path))
```
maciekgajewski, line 33: Is this possible? Is this check required? Shouldn't you rather use ```assert``` ?

course7/homeworks/waleed/ls/main.cc
```diff
@@ -1,16 +1,128 @@
...
+		ss << "/";
+
+	if (longArg)
+	{
+		ss << "\t" << fileSize(path);
+	}
+
+	ss << std::endl;
+
+	if (treeArg && bfs::is_directory(path))
```
maciekgajewski, line 54: Every call to filesystem functions is potentially expensive, as it requires talking to OS, and potentially doing I/O

maciekgajewski (approved):
Good. But I think that the program could be much more efficient, ie perform much less IO operations

course7/homeworks/waleed/ls/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -std=c++14")
 
 project (LS)
 
-find_package(Boost COMPONENTS filesystem system)
+find_package(Boost COMPONENTS filesystem program_options system)
```
maciekgajewski: I don't know which compiler are you using. but you should set CMAKE_CXX_STANDARD to 14, otherwise other compiles don't know that they should use C++-14 to compile the code.

PR #213, opened 2018-01-28
Homework 2: Extended ls with long and tree options

psde:
Demonstrating my inability to produce simple solutions for simple problems once more.

Complexity stems from the idea of fetching each file size only once and not iterate twice over one file.


maciekgajewski:
This is indeed an unnecessarily complex solution.
Theoretically, the only data structure you need to collect data about the tree is a stack.

I hereby challenge you: it is possible to use  only the implied stack to do it? You can collect the information and print in the same pass, to only call file_size once per file.

Simplicity!

PR #214, opened 2018-01-28
Homework 2

anton-kultsov:


course7/homeworks/anton/ls/main.cc
```diff
@@ -0,0 +1,135 @@
...
+        {
+            fs::path path(config.file_path);
+            print_file_tree(path, config);
+        }
+
+        return 0;
+    }
+    catch(std::exception& ex)
+    {
+        std::cout << "\033[1;31m" << ex.what() << "\033[0m" << std::endl;
```
maciekgajewski: Don't.
If you let the exception pass trough, a correct core dump file can be created that contains the state of the process at the moment of _throwing_ the exception, including the location of the throw, stacktrace, memory etc.

Otherwise it's impossible to debug the problem

course7/homeworks/anton/ls/main.cc
```diff
@@ -0,0 +1,135 @@
...
+
+configuration parse_args(int argc, char *argv[]);
+void print_file_tree(const fs::path& dir, const configuration& config, const std::string& filler = "");
+size_t get_file_size(const fs::path&);
+
+int main(int argc, char *argv[])
+{
+    try
+    {
+        const configuration& config = parse_args(argc, argv);
```
maciekgajewski: I don't see much point in using const ref here.
Returning stuff is cheap; an instance must exist anyway.

course7/homeworks/anton/ls/main.cc
```diff
@@ -0,0 +1,135 @@
...
+        po::variables_map vm;
+        po::store(po::command_line_parser(argc, argv).options(desc).positional(po).run(), vm);
+        po::notify(vm);
+
+        if (config.show_help)
+            print_help();
+    }
+    catch(boost::exception& ex)
+    {
+        print_help();
```
maciekgajewski: This complexity is unnecessary. See my other comment: just let the exceptions flow.

course7/homeworks/anton/ls/main.cc
```diff
@@ -0,0 +1,135 @@
...
+
+    po::options_description desc("Allowed options");
+    desc.add_options()
+        ("help,h", po::bool_switch(&config.show_help), "show help")
+        ("tree,t", po::bool_switch(&config.show_tree), "show tree")
+        ("long,l", po::bool_switch(&config.show_long), "show long")
+        ("file", po::value<std::string>(&config.file_path), "file path")
+    ;
+
+    auto print_help = [&]
```
maciekgajewski: Oh this could be just a simple function! no need for this.

course7/homeworks/anton/ls/main.cc
```diff
@@ -0,0 +1,135 @@
...
+}
+
+void print_file_info(const fs::path& file, const configuration& config, const std::string& filler)
+{
+    std::cout << filler;
+
+    if (config.show_long)
+    {
+        if (config.show_tree)
+            std::cout << std::setfill('.');
```
maciekgajewski: spaces would be better otherwise this will be unusable; selecting the size or the name with simple mouse double-click will select the entire line.

course7/homeworks/anton/ls/main.cc
```diff
@@ -0,0 +1,135 @@
...
+
+        std::cout << std::left << std::setw(12) << get_file_size(file);
+    }
+
+    std::cout << file.filename().string();
+
+    if (fs::is_directory(file))
+        std::cout << fs::path::preferred_separator;
```
maciekgajewski: standard posix ```/``` is our preferred separator :)

course7/homeworks/anton/ls/main.cc
```diff
@@ -0,0 +1,135 @@
...
+    std::cout << std::endl;
+}
+
+std::string get_increased_filler(const std::string filler)
```
maciekgajewski: you've forgot ```&``` in the param here, creating an extra copy

course7/homeworks/anton/ls/main.cc
```diff
@@ -0,0 +1,135 @@
...
+{
+    const static std::string& filler_diff = "    ";
```
maciekgajewski: It's OK to simply use string literal in such a case. Creating string instance is completely unnecessary here.

course7/homeworks/anton/ls/main.cc
maciekgajewski: Don't use color unless you make sure first that the output is a console. If the output is redirected to a file, or piped to another program for parsing, this will only create a mess.
When in doubt, don't use them at all.

maciekgajewski (approved):
Mostly correct, if bit too complex.

course7/homeworks/anton/ls/main.cc
anton-kultsov: my strategy was to use const refs whenever compiler swallows them without complaints))
I've just learned about RVO, is it what you were talking about?

course7/homeworks/anton/ls/main.cc
anton-kultsov: aah, my understanding of core dumps was completely wrong, thank you for the explanation. But, anyway, I believe, that program should not be core dumped if the user provided a wrong file name, or there is an issue with access rights, etc. So i changed `std::exception` to `boost::filesystem::error`. Is it ok?

anton-kultsov: fixed

course7/homeworks/anton/ls/main.cc
anton-kultsov: changed to function. but i don't like both variants. Possibly allowing the code duplication could be even better.

course7/homeworks/anton/ls/main.cc
anton-kultsov: the intent of this catch is to print help message if the user passes wrong/unknown parameters. Changed `boost::exception` to `boost::program_options::error`. is that acceptable?

course7/homeworks/anton/ls/main.cc
anton-kultsov: fixed

course7/homeworks/anton/ls/main.cc
anton-kultsov: changed. sounds like it is a matter of preference, though))

course7/homeworks/anton/ls/main.cc
anton-kultsov: lost my grip while refactoring))

course7/homeworks/anton/ls/main.cc
anton-kultsov: hm, another problem with one small refactoring. these two lines used to reside in `print_file_tree` where this `const static` had more sense (i hope it had)))

course7/homeworks/anton/ls/main.cc
```diff
@@ -0,0 +1,126 @@
...
+        po::store(po::command_line_parser(argc, argv).options(desc).positional(po).run(), vm);
+        po::notify(vm);
+
+        if (config.show_help)
+            print_help(fs::path(argv[0]).filename().string(), desc);
+    }
+    catch(po::error& ex)
+    {
+        std::cout << "Arguments error: " << ex.what() << std::endl;
+        print_help(fs::path(argv[0]).filename().string(), desc);
```
anton-kultsov, line 75: @maciekgajewski, if `fs::path(argv[0]).filename().string()` returned not `string&` but just `string` instance, would it have any sense to change?
`void print_help(const std::string& binary_name, const po::options_description& desc)` to 
`void print_help(const std::string binary_name, const po::options_description& desc)`

maciekgajewski: no. The most common use of function params it just to look at them. Using const ref gives you assurance that no unnecessary copy will be created, and that the actual object referenced by the reference will be alive until the end of you function.
Therefore: when in doubt, take (value) params by const reference.

course7/homeworks/anton/ls/main.cc
```diff
@@ -0,0 +1,126 @@
...
+        std::cout << std::left << std::setw(12) << get_file_size(file);
+
+    std::cout << file.filename().string();
+
+    if (fs::is_directory(file))
+        std::cout << '/';
+    std::cout << std::endl;
+}
+
+std::string get_increased_filler(const std::string& filler)
```
maciekgajewski, line 95: It would be cheaper to just stream the spaces to the stream in a loop, than to create a temporary string

maciekgajewski (approved):
Much cleaner now!

PR #215, opened 2018-01-30
Homework 2 - Patrick

Pvanhesteren:


course7/homeworks/patrickh/ls/ls.cc
```diff
@@ -1,14 +1,99 @@
 #include <iostream>
 #include <boost/filesystem.hpp>
+#include <boost/program_options.hpp>
```
maciekgajewski, line 3: It's always better to order the includes from the most general to the most specific. In this case the boost includes should be before iostream.

course7/homeworks/patrickh/ls/ls.cc
```diff
@@ -1,14 +1,99 @@
...
 
-int main()
+namespace fs = boost::filesystem;
+namespace po = boost::program_options;
+
+class file_summary
+{
+    public:
```
maciekgajewski, line 10: Just use ```struct``` and drop the ```public```

course7/homeworks/patrickh/ls/ls.cc
```diff
@@ -1,14 +1,99 @@
...
+        std::string name;
+        std::vector<file_summary> summaries;
+        size_t size;
+};
+
+std::string stringify_summary(const file_summary& summary, bool detailed)
```
maciekgajewski, line 16: Instead of appending to the string, you'd get more performant result using ```std::stringstream```

course7/homeworks/patrickh/ls/ls.cc
```diff
@@ -1,14 +1,99 @@
...
+        }
+    }
+    summary->name = fileName;
+    try
+    {
+        summary->size += fs::file_size(path);
+    }
+    catch (const fs::filesystem_error& err)
+    {
+        //catching some file size errors here
```
maciekgajewski, line 67: Print them to stderr

maciekgajewski (changes requested):
Quite clean solution, I like it except for the ```new```. Please, remove it!

PR #220, opened 2018-02-06
Complex nr calculator

MartSlot:
I probably missed some things somewhere, but basis seems to do what the assignment said? :)

### DIFF
--- a/course7/homeworks/mart/complex/complex.cc
+++ b/course7/homeworks/mart/complex/complex.cc
@@ -67,39 +67,44 @@ public:
 
 	void execute()
 	{
-		std::size_t i = 0;
-		while (i < line.length())
+		try
 		{
-			switch (line[i])
+			std::size_t i = 0;
+			while (i < line.length())
 			{
-			case '+':
-				add();
-				i++;
-				break;
-			case '-':
-				subtract();
-				i++;
-				break;
-			case '*':
-				multiply();
-				i++;
-				break;
-			case '/':
-				divide();
-				i++;
-				break;
-			default:
-				i = parseAndPushComplex(i);
+				switch (line[i])
+				{
+				case '+':
+					add();
+					i++;
+					break;
+				case '-':
+					subtract();
+					i++;
+					break;
+				case '*':
+					multiply();
+					i++;
+					break;
+				case '/':
+					divide();
+					i++;
+					break;
+				default:
+					i = parseAndPushComplex(i);
+				}
+
+				while (i < line.length() && line[i] == ' ')
+					i++;
 			}
 
-			while (i < line.length() && line[i] == ' ')
-				i++;
-		}
-
-		if (stack.empty())
-			std::cout << "empty" << std::endl;
-		else
 			std::cout << stack.top() << std::endl;
+		}
+		catch(const std::exception& e)
+		{
+			std::cerr << "Error executing line: " << e.what() << std::endl;
+			std::cout << "empty" << std::endl;
+		}
 	}
 
 private:
--- a/course7/homeworks/mart/complex/complex.cc
+++ b/course7/homeworks/mart/complex/complex.cc
@@ -21,9 +21,25 @@ struct Complex
 		return { real + rhs.real, imaginary + rhs.imaginary };
 	}
 
-	Complex operator+(double rhs) const
+	Complex operator-(const Complex& rhs) const
 	{
-		return { real + rhs, imaginary };
+		return { real - rhs.real, imaginary - rhs.imaginary };
+	}
+
+	Complex operator*(const Complex& rhs) const
+	{
+		return {
+			(real * rhs.real) - (imaginary * rhs.imaginary),
+			(real * rhs.imaginary) + (imaginary * rhs.real)
+		  };
+	}
+
+	Complex operator/(const Complex& rhs) const
+	{
+		return {
+			((real * rhs.real) + (imaginary * rhs.imaginary)) / ((rhs.real * rhs.real) + (rhs.imaginary * rhs.imaginary)),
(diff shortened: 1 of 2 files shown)


course7/homeworks/mart/complex/complex.cc
```diff
@@ -0,0 +1,182 @@
...
+		auto pair = popPair();
+		stack.push(pair.second / pair.first);
+	}
+
+private:
+	std::string line;
+	std::stack<Complex> stack;
+};
+
+int main(/*int argc, char** argv*/)
```
MartSlot: Mathias already notified me I should uncomment this while removing the args and arg names
Also I should return 0

course7/homeworks/mart/complex/complex.cc
```diff
@@ -0,0 +1,182 @@
...
+#include <string>
+#include <stack>
+#include <regex>
+
+struct Complex
+{
+	Complex() = default;
+	Complex(double r) : real(r) {}
+	Complex(double r, double i) : real(r) , imaginary(i) {}
+	~Complex() = default;
```
maciekgajewski: This is not required. destructor is always generated.
The purpose of ```= default``` is to ask compiler to generate the method, if otherwise it would refrained for it.

course7/homeworks/mart/complex/complex.cc
```diff
@@ -0,0 +1,182 @@
...
+			((real * rhs.real) + (imaginary * rhs.imaginary)) / ((rhs.real * rhs.real) + (rhs.imaginary * rhs.imaginary)),
+			((imaginary * rhs.real) - (real * rhs.imaginary)) / ((rhs.real * rhs.real) + (rhs.imaginary * rhs.imaginary))
+		  };
+	}
+
+	double real = 0.0;
+	double imaginary = 0.0;
+};
+
+bool operator==(Complex lhs, Complex rhs)
```
maciekgajewski: Why have you decided to implement this operator as a free function, if other operators are methods? Why the inconsistency?

course7/homeworks/mart/complex/complex.cc
maciekgajewski: This is not true at all.
```main``` is special. It is OK not to take params, and it is also OK not to have a ```return``` statement in there. (0 is then the implied return value)

File: course7/homeworks/mart/complex/complex.cc [hunk 4]
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+private:
+	std::string line;
+	std::stack<Complex> stack;
+};
+
+int main(/*int argc, char** argv*/)
+{
+	for (std::string line; std::getline(std::cin, line); ) {
```
maciekgajewski: I was expecting that there would be one token per line:
```
2
+
2+i2
```

course7/homeworks/mart/complex/complex.cc
```diff
@@ -0,0 +1,182 @@
...
+class Calculator
+{
+public:
+	Calculator(const std::string& l) : line(l) {}
+
+	void execute()
+	{
+		try
+		{
+			std::size_t i = 0;
```
maciekgajewski: Drop the parsing, just expect one token per line. just like ```dc```.

maciekgajewski (changes requested):
Quite nice! Pleasem simplify the parsing; assume one token per line

course7/homeworks/mart/complex/complex.cc
MartSlot: Thx, fixed

course7/homeworks/mart/complex/complex.cc
MartSlot: Ah yes, I took it straight from the meeting03 code but forgot to fix this one

course7/homeworks/mart/complex/complex.cc
MartSlot: Ah, that makes it a lot simpler indeed, nice! Fixed!

course7/homeworks/mart/complex/complex.cc
```diff
@@ -0,0 +1,155 @@
...
+
+	Complex operator/(const Complex& rhs) const
+	{
+		return {
+			((real * rhs.real) + (imaginary * rhs.imaginary)) / ((rhs.real * rhs.real) + (rhs.imaginary * rhs.imaginary)),
+			((imaginary * rhs.real) - (real * rhs.imaginary)) / ((rhs.real * rhs.real) + (rhs.imaginary * rhs.imaginary))
+		  };
+	}
+
+	bool operator==(const Complex& rhs)
```
maciekgajewski, line 47: This should be const

course7/homeworks/mart/complex/complex.cc
```diff
@@ -0,0 +1,155 @@
...
+				else if (line == "*")
+					multiply();
+				else if (line == "/")
+					divide();
+				else
+					parseAndPushComplex(line);
+			}
+			catch(const std::exception& e)
+			{
+				std::cerr << "Error executing line: " << e.what() << std::endl;
```
maciekgajewski, line 84: I think that after an error the calculator would be in an inconsistent state, so the error should be fatal.
Therefore, it would be better to let the exception fly trough.

course7/homeworks/mart/complex/complex.cc
```diff
@@ -0,0 +1,155 @@
...
+				std::cout << "empty" << std::endl;
+			else
+				std::cout << stack.top() << std::endl;
+		}
+	}
+
+private:
+	void parseAndPushComplex(const std::string& line)
+	{
+		static const std::regex regex("^([0-9]+.[0-9]+)(?:\\+i([0-9]+.[0-9]+))?$");
```
maciekgajewski, line 96: I think the parsing should be part of the complex class.

maciekgajewski (approved):
minor comments only

PR #221, opened 2018-02-07
Homework 3: Complex number RPN calculator

psde:


course7/homeworks/mathias/complex_rpn/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+project(complex_rpn VERSION 1.0 LANGUAGES CXX)
+
+if(MSVC)
+    add_compile_options("/permissive-" "/W4" "/WX" "$<$<CONFIG:RELEASE>:/O2>")
+else()
+    add_compile_options("-std=c++14" "-Wall" "-Wextra" "-Werror" "$<$<CONFIG:RELEASE>:-O3>")
+endif()
+
+add_executable(complex_rpn main.cpp complex.h complex.cpp rpn_calculator.h rpn_calculator.cpp)
+target_include_directories(complex_rpn PRIVATE ${Boost_INCLUDE_DIR})
```
maciekgajewski, line 12: Boost_INCLUDE_DIR is not going to be populated, if there is no find_package(Boost ...)

course7/homeworks/mathias/complex_rpn/complex.cpp
```diff
@@ -0,0 +1,61 @@
...
+		throw std::runtime_error("Input value too big to store.");
+	}
+}
+
+Complex Complex::operator+(const Complex& rhs)
+{
+	return Complex { mReal + rhs.mReal, mImaginary + rhs.mImaginary };
+}
+
+Complex Complex::operator-(const Complex& rhs)
```
maciekgajewski, line 38: It is better to place functions this size in the header file, so that compiler can see them and inline them when compiling other compilation units.

course7/homeworks/mathias/complex_rpn/complex.h
```diff
@@ -0,0 +1,25 @@
...
+class Complex
+{
+public:
+	Complex(double real, double imaginary);
+	explicit Complex(const std::string& str);
+
+	Complex operator+(const Complex& rhs);
+	Complex operator-(const Complex& rhs);
+	Complex operator*(const Complex& rhs);
+	Complex operator/(const Complex& rhs);
```
maciekgajewski, line 15: Shouldn't these be const?

course7/homeworks/mathias/complex_rpn/main.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <iostream>
+#include <string>
+
+int main(int , char** )
+{
+	RpnCalculator calc;
+	std::string line;
+
+	std::cout << "> ";
```
maciekgajewski, line 12: Prompt is cool, but useless, and even damaging in batch mode, when commands are read from file. Then the prompt just creates a noise in the output stream. Imagine that you want to use the calculator as part of some automated ipeline, when input is generated and output parsed.

course7/homeworks/mathias/complex_rpn/main.cpp
```diff
@@ -0,0 +1,30 @@
...
+	while (std::getline(std::cin, line))
+	{
+		try
+		{
+			calc.parse(line);
+		}
+		catch (const std::runtime_error& e)
+		{
+			std::cout << "Error: " << e.what() << std::endl;
```
maciekgajewski, line 21: 1. If the error is not recoverable, then let the exception flow out and kill the process. In this case, any error makes the state inconsistent, as any subsequent commands can make assumptions/expectations regarding the state of the stack.
2. If errors are recoverable, hen print them to STDERR (```std::cerr```) to not pollute the output. When writing a Unix tool, always expect that someone may want to parse your output.

maciekgajewski (approved):
Nice, simple :)
Minor errors around output, error handling and const-correctness.

PR #222, opened 2018-02-07
Meeting 3: Complex numbers RPN calculator

fredpantuzza:


course7/homeworks/fred/complex-rpn/calculator.cc
```diff
@@ -0,0 +1,64 @@
...
+      case Operator::SUBTRACTION:
+        return operand1 - operand2;
+
+      case Operator::MULTIPLICATION:
+        return operand1 * operand2;
+
+      case Operator::DIVISION:
+        return operand1 / operand2;
+
+      default:
```
maciekgajewski: If you don't use the 'default' label, then the compiler will warn you if you don't use all the values of the enum.

course7/homeworks/fred/complex-rpn/complex.h
```diff
@@ -0,0 +1,21 @@
...
+  double imaginary;
+
+  Complex operator+ (const Complex& other) const;
+  Complex operator- (const Complex& other) const;
+  Complex operator* (const Complex& other) const;
+  Complex operator/ (const Complex& other) const;
+};
+
+template <typename Stream>
+Stream& operator<< (Stream& stream, const Complex& c) {
```
maciekgajewski: This should be the other way around:
- the ostream operator, which is not perf critical and is rather large, should be in the .cc file.
- the small arithmetical operations, that would benefit from inlining, should be defined here in the header.

course7/homeworks/fred/complex-rpn/main.cc
```diff
@@ -0,0 +1,65 @@
...
+  std::cout << "Result so far: " << result << std::endl;
+}
+
+int main(int argc, char** args) {
+  std::cout << "Welcome!" << std::endl;
+  std::cout << "Type 'q' to quit" << std::endl << std::endl;
+
+  RpnCalculator calculator;
+
+  std::string input = "";
```
maciekgajewski: There is no point initializing the value to ```""```. The default-constructed ```std::string``` is empty.
Have you read the documentation of std::string? If not, do it!

course7/homeworks/fred/complex-rpn/main.cc
```diff
@@ -0,0 +1,65 @@
...
+  while (true) {
+    std::getline(std::cin, input);
+    if (input == "q")
+      break;
+
+    bool parseFailed = false;
+
+    try {
+      // try to parse a number first
+      auto parseComplexResult = tryParseComplexFromString(input);
```
maciekgajewski: The code would be much clearer if you'd have a function like ```isOperaror```.
The 'tryParse' functions do 2 things. It is always a good idea for function/class to do exactly 1 thing.

course7/homeworks/fred/complex-rpn/main.cc
```diff
@@ -0,0 +1,65 @@
...
+        else {
+          parseFailed = true;
+        }
+      }
+    }
+    catch (const std::runtime_error& e) {
+      std::cout << e.what() << std::endl;
+    }
+
+    if (parseFailed) {
```
maciekgajewski: Here you can see the cost of the 'tryParse' approach: you have, competing methods of reporting errors: by exception and by return value, and you need to have two separate error handlers.

course7/homeworks/fred/complex-rpn/utils.h
```diff
@@ -0,0 +1,66 @@
+#include "complex.h"
+#include "calculator.h"
+
+#include <string>
+#include <stdexcept>
+#include <utility>
+
+Complex parseComplexFromString(const std::string& s) {
```
maciekgajewski: Don't you thing this logic should be part of the ```Complex``` class?
This is what a user would expect from a complex number library.

course7/homeworks/fred/complex-rpn/utils.h
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  return Complex{ realPart, imaginaryPart };
+}
+
+std::pair<bool, Complex> tryParseComplexFromString(const std::string& s) {
+  try {
+    auto result = parseComplexFromString(s);
+    return std::make_pair(true, result);
+  }
```
maciekgajewski: Throwing exceptions in C++ is expensive, and shouldn't be use as a part of normal logic flow.

course7/homeworks/fred/complex-rpn/complex.cc
```diff
@@ -0,0 +1,42 @@
...
+  Complex result;
+  result.real = (this->real * other.real) - (this->imaginary * other.imaginary);
+  result.imaginary = (this->imaginary * other.real) + (this->real * other.imaginary);
+  return result;
+}
+
+Complex Complex::operator/ (const Complex& other) const
+{
+  if (other.real == 0.0 && other.imaginary == 0.0) {
+    // TODO check with epsilon?
```
maciekgajewski: No. If comparing doubles doesn't use epsilons, why would comparing Complexes would?

course7/homeworks/fred/complex-rpn/complex.cc
```diff
@@ -0,0 +1,42 @@
...
+    throw std::runtime_error("Division by zero");
```
maciekgajewski: This violates the "pay only fr what you are using" principle. Make this a narrow contract function, and make this user's responsibility to ensure the values are valid.

course7/homeworks/fred/complex-rpn/main.cc
```diff
@@ -0,0 +1,65 @@
+#include "complex.h"
+#include "calculator.h"
+#include "utils.h"
+
+#include <iostream>
+#include <stdexcept>
+
+void insert(RpnCalculator& calculator, Complex complex) {
```
maciekgajewski, line 8: What is the value added by this function?

maciekgajewski (changes requested):
The flow in 'main' could be so much simpler!

course7/homeworks/fred/complex-rpn/complex.h
fredpantuzza: What if I add the word 'inline' to the arithmetical operations definitions? Would that have the same effect as putting the implementation in the header file?

maciekgajewski: It won't help at all. To inline the function, compiler has to access the function's definition when compiling calculator.cc, so it has to be in the included header.

course7/homeworks/fred/complex-rpn/main.cc
fredpantuzza: I just wanted to maintain a standard

PR #224, opened 2018-02-10
Implement homework 3 with some additions

patrickkostjens:
Aside from the basics for homework 3, this also implements q (to quit), c (to clear the stack), and negative numbers.

### DIFF
--- a/course7/homeworks/patrickkostjens/complex/complex.cc
+++ b/course7/homeworks/patrickkostjens/complex/complex.cc
@@ -1,0 +1,133 @@
+#include <cmath>
+#include <iostream>
+#include <stack>
+#include <stdlib.h>
+#include <string>
+#include <regex>
+
+struct Complex
+{
+	double real = 0.0;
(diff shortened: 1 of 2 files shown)


course7/homeworks/patrickkostjens/complex/complex.cc
```diff
@@ -0,0 +1,133 @@
...
+struct Complex
+{
+	double real = 0.0;
+	double imaginary = 0.0;
+
+	Complex() = default;
+	Complex(double r) : real(r) {}
+	Complex(double r, double i) : real(r), imaginary(i) {}
+
+	Complex(const Complex& other) : real(other.real), imaginary(other.imaginary) {}
```
maciekgajewski, line 17: Do you need to implement this? Can you just use the implementation provided by the compiler? (applies also to ```operator=```)

maciekgajewski (approved):
Good! Simple, I like simple.

PR #225, opened 2018-02-10
homework 3 - Patrick

Pvanhesteren:


course7/homeworks/patrickh/homework3/calculator.cpp
```diff
@@ -0,0 +1,73 @@
...
+    if (input == "+")
+        add();
+    else if (input == "-")
+        subtract();
+    else if (input == "*")
+        multiply();
+    else if (input == "/")
+        divide();
+    else
+        mStack.emplace(input);
```
patrickkostjens, line 16: I think it's more clear to explicitly use the Complex constructor here, instead of replying on an implicit cast.

course7/homeworks/patrickh/homework3/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+cmake_minimum_required(VERSION 3.0.0)
+set(CMAKE_CXX_STANDARD 14)
+project(calculator)
+
+include_directories(${Boost_INCLUDE_DIRS})
```
maciekgajewski, line 5: This is not going to do much. The ``Boost_*``` variables are set when you execute ```find_package(Boost ...)```, which you don't

course7/homeworks/patrickh/homework3/calculator.cpp
```diff
@@ -0,0 +1,73 @@
...
+void Calculator::divide()
+{
+    if(mStack.size() < 2)
+        throw std::runtime_error("Not enough values to be divided");
+    Complex c1 = mStack.top();
+    mStack.pop();
+    Complex c2 = mStack.top();
+    mStack.pop();
+    mStack.emplace(c1 / c2);
+}
```
maciekgajewski, line 61: This code here is correct but very verbose. Please, have a look at Patrick K's solution, he managed to squeeze it all in some 8 lines of code.

course7/homeworks/patrickh/homework3/complex.cpp
```diff
@@ -0,0 +1,55 @@
...
+{
+    return Complex(this->mReal + other.mReal, this->mImaginary + other.mImaginary);
+}
+
+Complex Complex::operator-(const Complex& other)
+{
+    return Complex(this->mReal - other.mReal, this->mImaginary - other.mImaginary);
+}
+
+Complex Complex::operator*(const Complex& other)
```
maciekgajewski, line 27: It would be better if these small functions were implemented in the header file. Easier to read by humans, and easier to optimize by the compiler.

course7/homeworks/patrickh/homework3/main.cpp
```diff
@@ -0,0 +1,20 @@
...
+    std::string line;
+    while(std::getline(std::cin, line))
+    {
+        try
+        {
+            calculator.parse(line);
+        }
+        catch(const std::runtime_error& error)
+        {
+            std::cout << "Some error occured: " << error.what() << std::endl;
```
maciekgajewski, line 16: Maybe use ```cerr``` instead of ```cout```

maciekgajewski (commented):
Good, but I don't like that you've mixed two assignments into single PR. There are separate assignments.

PR #226, opened 2018-02-11
hw3&4: dc with complex numbers

anton-kultsov:
Examples of usage:
```bash
$ echo "1+i1p" | ./dc
1+i1
$ echo "_1-i1p" | ./dc
-1-i1
$ echo "__if" | ./dc
-i1
-0
$ echo "ii*p" | ./dc
-1
```

### DIFF
--- a/course7/homeworks/anton/dc/complex.h
+++ b/course7/homeworks/anton/dc/complex.h
@@ -1,0 +1,48 @@
+#include <iostream>
+
+struct Complex
+{
+    Complex()
+    {
+
+    }
+
+    Complex(double r, double i)
+    {
+        mR = r;
+        mI = i;
+    }
+
+    Complex operator+(const Complex& rhs)
+    {
+        return Complex(mR + rhs.mR, mI + rhs.mI);
+    }
+
+    Complex operator-(const Complex& rhs)
+    {
+        return Complex(mR - rhs.mR, mI - rhs.mI);
+    }
+
+    Complex operator*(const Complex& rhs)
+    {
+        return Complex(mR*rhs.mR - mI*rhs.mI, mR*mI + rhs.mR*rhs.mI);
+    }
+
+    Complex operator/(const Complex& rhs)
+    {
+        double denomenator = rhs.mI*rhs.mR + rhs.mI*rhs.mI;
+        return Complex((mR*rhs.mR - mI*rhs.mI)/denomenator, (mI*rhs.mR - mR*rhs.mI)/denomenator);
+    }
+
+    bool is_real()
+    {
+        return mI == 0.0;
+    }
+
+    double mR = 0.0;
+    double mI = 0.0;
+};
+
+std::ostream& operator<<(std::ostream& s, const Complex& rhs);
+
+std::istream& operator>>(std::istream& s, Complex& rhs);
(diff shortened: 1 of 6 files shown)


course7/homeworks/anton/dc/complex.h
```diff
@@ -0,0 +1,63 @@
...
+    bool is_real()
+    {
+        return mI == 0.0;
+    }
+
+    double mR = 0.0;
+    double mI = 0.0;
+};
+
+typedef Complex (*operation)(Complex&, Complex&);
```
maciekgajewski, line 50: Prefer modern ```using``` instead of old-school ```typedef```

course7/homeworks/anton/dc/complex_stack.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
+
+#include "complex.h"
+#include "stack"
```
maciekgajewski, line 4: include ```<stack>```. Use angular brackets for 3rd party code, and quotes for local includes. Mostly by conventions these days.

course7/homeworks/anton/dc/complex.h
```diff
@@ -0,0 +1,63 @@
...
+
+Complex add(Complex& n1, Complex& n2);
+
+Complex substruct(Complex& n1, Complex& n2);
```
maciekgajewski, line 54: *substract

course7/homeworks/anton/dc/complex_stack.cc
```diff
@@ -0,0 +1,59 @@
...
+    {
+        complex::Complex number;
+        std::cin >> number;
+        stack.push(number);
+        return true;
+    }
+    return false;
+}
+
+void ComplexStack::exec_operation(complex::operation op)
```
maciekgajewski, line 41: This is completely unnecessary, right? There is no need for dynamic dispatch if you're coming from a static ```switch``` in main.
IMO, unneeded complexity is the worst of bugs.

course7/homeworks/anton/dc/complex_stack.h
```diff
@@ -0,0 +1,50 @@
...
+
+
+class ComplexStack
```
maciekgajewski, line 7: Make sure that your classes ans functions do one thing, and one thing only.
What is this class doing? Is it a container? A command dispatcher? A parser?

maciekgajewski (commented):
It is a correct implementation, but oh so overly complex.
1. The dynamic routing trough function pointers - not needed, but adds tens of lines of code. I see how this could be base for the future dynamic loading implementation, you shouldn't add complexity unless it's needed.
2. General design: it is difficult to say which part of code is responsible for what.

course7/homeworks/anton/dc/complex_stack.cc
anton-kultsov: agree with you partially, it is not *completely* unnecessary since it makes this class a bit more extendable and reusable and allows to implement your next requirement of using DLs, right?

course7/homeworks/anton/dc/complex_stack.h
anton-kultsov: in terms of C# this class contains a bunch of extension methods for `std::stack<complex::Complex>`. I definitely don't like it. Would it be better if i implemented it as a set of functions taking `std::stack<complex::Complex>` as the first parameter? Formally, instead of one class breaking SRP, we would have a few functions compliant to SRP, but would it make any difference?

PR #227, opened 2018-02-13
Complex calculator

waleedlotfy:


course7/homeworks/waleed/complex/calculator.cpp
```diff
@@ -0,0 +1,52 @@
...
+    if (str.size() == 1) // Operator
+    {
+        if (st.size() < 2)
+        {
+            throw std::runtime_error("Not enough operands to perform operation");
+        }
+        Complex operand2 = st.top(); st.pop();
+        Complex operand1 = st.top(); st.pop();
+        Complex newTop;
+        switch (str[0])
```
patrickkostjens, line 16: Maybe print a message in case of an unrecognized operator, so that it's clear what happened.

course7/homeworks/waleed/complex/complex.h
```diff
@@ -0,0 +1,32 @@
...
+    double GetR() const;
+
+    double GetI() const;
+
+private:
+    double mR;
+    double mI;
+};
+
+inline std::ostream& operator<<(std::ostream& os, const Complex& arg)
```
maciekgajewski, line 28: It should be the other way around. The complex and slow operator<< should go to the C++ file, while the small, most benefiting from inlining operators should be here.

course7/homeworks/waleed/complex/main.cpp
```diff
@@ -0,0 +1,27 @@
...
+    while (std::getline(std::cin, inp))
+    {
+        calc.PushLine(inp);
+    }
+
+    try
+    {
+        std::cout << calc.GetTop() << std::endl;
+    }
+    catch (std::exception)
```
maciekgajewski, line 21: It is not a good idea to use exceptions for control flow, as they are horribly slow to throw in C++.

Maybe you could design the API of ```GetTop``` so that it returns a pointer to Complex, or null if stack is empty?

course7/homeworks/waleed/complex/calculator.cpp
maciekgajewski: Or even better: throw and exception

maciekgajewski (approved):
Nice, simple.
The error handling could be better

course7/homeworks/waleed/complex/main.cpp
waleedlotfy: I agree on the performance issue for exceptions, but the problem with using pointer to fix it is that if I return address of an element in the Calculator's stack, it might be changed in the future (by pops and pushes) without the holder of that address noticing. 
And if I create a new instance in the heap and return its address, who should be responsible for destroying it?

PR #228, opened 2018-02-17
Homework05: Split up into separate files and implement plugin system

patrickkostjens:
I had some issues with resource management/memory allocation with the plugin system. Iterating over the files twice is done to prevent the function from being overwritten at every iteration thereby making only the last plugin function properly. I'm not sure (yet) how to improve/fix this in a better way.


course7/homeworks/patrickkostjens/complex/calculator.hh
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "complex.hh"
+
+#include <iostream>
+#include <map>
+#include <stack>
+#include <string>
+
+typedef void* (*plugin)(std::stack<Complex>*);
```
maciekgajewski: Prefer modern ```using``` over old-school ```typedef```

maciekgajewski: Use references instead of pointer, unless you want to create a semantics where passing null pointer means something.

course7/homeworks/patrickkostjens/complex/main.cc
```diff
@@ -0,0 +1,44 @@
...
+
+int main(int argc, char** argv)
+{
+	Calculator c;
+	int plugin_count = 0;
+	for (const auto& it : fs::directory_iterator(fs::path(".")))
+		if (it.path().extension() == ".so" || it.path().extension() == ".dylib")
+			plugin_count++;
+
+	plugin *plugins = new plugin[plugin_count];
```
maciekgajewski: You already store the function pointers in the calculator. Do you need to store them separately here?

maciekgajewski: The "new" again...
Never use it, unless you are working on some low-level stuff. And then, if you do, every "new" should be paired with "delete".
If you need variable-sized storage, use ```std::vector```

course7/homeworks/patrickkostjens/complex/CMakeLists.txt
```diff
@@ -2,4 +2,13 @@ cmake_minimum_required(VERSION 3.0)
...
+find_package(Boost REQUIRED COMPONENTS filesystem)
+
+INCLUDE_DIRECTORIES(${Boost_INCLUDE_DIR})
+
+add_library(print SHARED print.cc complex.cc complex.hh)
+SET_TARGET_PROPERTIES(print PROPERTIES PREFIX "")
+
+add_executable(complex main.cc calculator.hh calculator.cc complex.hh complex.cc)
+
+TARGET_LINK_LIBRARIES(complex LINK_PUBLIC ${Boost_LIBRARIES})
```
maciekgajewski: On Linux, you'd have to link with "ld". You may want to practice detecting OS in CMake

maciekgajewski (changes requested):
Nice, simple, I like it!
Just get rid of this "new"

course7/homeworks/patrickkostjens/complex/calculator.hh
patrickkostjens: Fixed.

course7/homeworks/patrickkostjens/complex/main.cc
patrickkostjens: Before, it broke when I didn't do it, but I now find another way to do it where I don't need to store them here. Hopefully that's better. It at least looks simpler.

patrickkostjens: Removed. I'm still getting used to `new` not being needed basically.

course7/homeworks/patrickkostjens/complex/CMakeLists.txt
patrickkostjens: I'm not sure what you mean by this.

patrickkostjens:
PR should be fixed and clean now.

course7/homeworks/patrickkostjens/complex/main.cc
```diff
@@ -3,27 +3,31 @@
...
+			if (!module)
+			{
+				std::cerr << "Error loading module " << name << " : " << ::dlerror() << std::endl;
+				return 1;
+			}
+			plugin p = reinterpret_cast<plugin>(::dlsym(module, "execute"));
+			if (!p)
+			{
+				std::cerr << "Symbol not found from file " << name << ", : " << ::dlerror() << std::endl;
+				return 1;
```
maciekgajewski, line 28: I'm not sure these returns should be here; they make the errors fatal. I don't think a failure to load a plugin should be fatal.

maciekgajewski (approved):
This is it!

PR #229, opened 2018-02-19
hw5: dc with plugins

anton-kultsov:
To make it simpler, i've removed the support for infix literals, like `1+i2`, postfix form to be used instead: `1i2+`


course7/homeworks/anton/dc/complex.h
```diff
@@ -34,7 +32,7 @@ struct Complex
 
     Complex operator/(const Complex& rhs)
     {
-        double denomenator = rhs.mI*rhs.mR + rhs.mI*rhs.mI;
+        double denomenator = rhs.mR*rhs.mR + rhs.mI*rhs.mI;
```
maciekgajewski, line 35: *denominator

course7/homeworks/anton/dc/operation.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
```
maciekgajewski, line 1: The header files are not needed.

maciekgajewski (approved):
Good!

PR #231, opened 2018-02-22
Homework 04: dc like interface

psde:
I added a few commands to be dc-like, and fixes the issues from the last PR.


maciekgajewski (approved):
Nice; I like it. Simple, easy to read.

PR #233, opened 2018-02-24
Homework 6: use own implementation of complex stack instead of std::stack

patrickkostjens:


course7/homeworks/patrickkostjens/complex/CMakeLists.txt
```diff
@@ -6,9 +6,9 @@ find_package(Boost REQUIRED COMPONENTS filesystem)
 
 INCLUDE_DIRECTORIES(${Boost_INCLUDE_DIR})
 
-add_library(print SHARED print.cc complex.cc complex.hh)
+add_library(print SHARED print.cc complex.cc complex.hh complex_stack.cc complex_stack.hh)
```
maciekgajewski: You could create a small static library with complex and stack, and then link both the executable and the plugin with it.
This would be the interface library you'd distribute to plugin writers.

course7/homeworks/patrickkostjens/complex/complex_stack.hh
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include "complex.hh"
+
+#include <memory>
+
+struct ComplexStackNode
```
maciekgajewski: The node is an implementation detail of the stack. It should be a nested, private class of ```ComplexStack```

course7/homeworks/patrickkostjens/complex/complex_stack.cc
```diff
@@ -0,0 +1,27 @@
+#include "complex_stack.hh"
+
+bool ComplexStack::empty() const
```
maciekgajewski: This one is so small,you could put it in the header, so that it can be inlined.

maciekgajewski (approved):
Very good!

patrickkostjens:
I incorporated the requested changes. It's indeed better this way.

maciekgajewski (approved):

PR #235, opened 2018-03-06
Homework 7

MartSlot:


course7/homeworks/mart/complex/complex_linked_list.cc
```diff
@@ -0,0 +1,33 @@
...
+{
+	return mSize == 0;
+}
+
+std::size_t ComplexLinkedList::size() const
+{
+	return mSize;
+}
+
+Complex& ComplexLinkedList::back()
```
maciekgajewski, line 13: these one-lines should really be in the header file. Such a small functions benefit from inlining the most.

maciekgajewski (approved):
Nice. Only the one-liners should live in the header.

PR #236, opened 2018-03-18
Homework 6 + 7

Pvanhesteren:


Pvanhesteren:
I'm still a little behind with the homeworks, but I'll catch up :)

course7/homeworks/patrickh/complex/calculator.cpp
```diff
@@ -0,0 +1,73 @@
...
+    if (input == "+")
+        add();
+    else if (input == "-")
+        subtract();
+    else if (input == "*")
+        multiply();
+    else if (input == "/")
+        divide();
+    else
+        mStack.push(input);
```
maciekgajewski, line 16: I don't think that allowing for implicit string->number conversion is a good API design.

course7/homeworks/patrickh/complex/calculator.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+void Calculator::divide()
+{
+    if(mStack.size() < 2)
+        throw new std::runtime_error("Not enough elements on the stack");
+    Complex c1 = mStack.top();
+    mStack.pop();
+    Complex c2 = mStack.top();
+    mStack.pop();
```
maciekgajewski, line 59: I see repeated code, and hence the opportunity to factor-out the commonalities

course7/homeworks/patrickh/complex/complex.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+#include<string>
+class Complex
+{
+public:
+    Complex(double real, double imaginary);
+    Complex(const std::string& input);
```
maciekgajewski, line 7: this should be explicit at least, IMO, or even a named constructor: ```static Complex fromString(const std::string&);```

course7/homeworks/patrickh/complex/complex_linked_list_node.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+#include "complex.h"
+#include "memory"
+struct ComplexLinkedListNode
```
maciekgajewski, line 4: This is a private class, an implementation detail. It doesn't deserve to be a top-level class, not to mention it's own source file. IMO this should be a private nested class of the list.

course7/homeworks/patrickh/homework3/calculator.cpp
```diff
@@ -13,56 +13,56 @@ void Calculator::parse(const std::string& input)
...
     else
-        mStack.emplace(input);
+        mStack.push(input);
 }
 
 void Calculator::add()
 {
     if(mStack.size() < 2)
-        throw std::runtime_error("Not enough values to be added");
+        throw new std::runtime_error("Not enough elements on the stack");
```
maciekgajewski, line 22: WRONG! you want to throw an exception, not a pointer to exception.

maciekgajewski (commented):
Good! I question few design choices, but other than that I like it.

PR #237, opened 2018-03-24
Implement chat client in C++, chat server in Python

patrickkostjens:
Some cleanup can still be done, but both the client (C++) and server (Python) work.


psde:
Very nice! I used your client to test my server, and also changed my protocol to be compatible to yours.

course7/homeworks/patrickkostjens/chat/chat_client.h
```diff
@@ -0,0 +1,32 @@
...
+    void on_message(websocketpp::connection_hdl, client::message_ptr msg);
+    void send_json(const json& obj);
+    void handle_handshake_reply(const json& obj);
+
+    client m_endpoint;
+    websocketpp::lib::thread m_thread;
+    websocketpp::connection_hdl m_hdl;
+
+    bool m_open;
+    bool m_is_registered;
```
psde: These are not initialized and therefore did not work on my machine when testing.

course7/homeworks/patrickkostjens/chat/chat_client.cc
```diff
@@ -0,0 +1,122 @@
...
+void ChatClient::handle_handshake_reply(const json& message)
+{
+    if (message.find("user_list") != message.end())
+    {
+        m_is_registered = true;
+        std::cout << "< Present users: ";
+        for (int i = 0; i < message["user_list"].size(); ++i)
+        {
+            if (i != 0) std::cout << ", ";
+            std::cout << message["user_list"][i]["name"].get<std::string>();
```
psde, line 102: I was surprised by this, as I implemented it as an array of strings (`"user_list": ["mathias", "patrickk", "maciek"]`), instead of an array of objects. Also you are using `name` in the handshake and `username` in the messages

patrickkostjens: I did that to account for the extension where extra user data is available. It would be kind of odd to have that in a separate list, so to keep compatibility with that extension, I figured having objects for this might be easier to deal with.

course7/homeworks/patrickkostjens/chat/chat_client.h
patrickkostjens: Good point, not sure why they didn't cause me problems, but I'll fix it.

